PR #20, opened 2019-10-29
Mapped d2d1helper.h and most of d2d1_1helper.h

nathan-alden-sr:
I mapped `d2d1helper.h` and most of `d2d1_1helper.h` with work consisting of the following:

- Constructors for several structs
- `public static readonly` default instances for several structs
- `IEquatable<T>` implemented for many structs
- Refactored the `ColorF` class into `DXGI_RGBA` plus a `ColorF` enum

I did *not* map the `SetDpiCompensatedEffectInput` function in `d2d1_1helper.h` since I'm not sure where it belongs or if it's valuable to map.


sources/Interop/DCommon/um/dcommon/D2D_POINT_2U.Manual.cs
```diff
@@ -0,0 +1,30 @@
...
+// Ported from um/d2d1helper.h in the Windows SDK for Windows 10.0.18362.0
+// Original source is Copyright © Microsoft. All rights reserved.
+
+using System;
+
+namespace TerraFX.Interop
+{
+    public partial struct D2D_POINT_2U : IEquatable<D2D_POINT_2U>
+    {
+        public static readonly D2D_POINT_2U DEFAULT = new D2D_POINT_2U(0, 0);
```
tannergooding: nit: It's probably not worth defining `DEFAULT` when:
* The header file doesn't itself define an explicit default
-and
* The constructor parameter defaults are the same as `new T()`

sources/Interop/DCommon/um/dcommon/D2D_POINT_2U.Manual.cs
```diff
@@ -0,0 +1,30 @@
...
+            this.y = y;
+        }
+
+        public bool Equals(D2D_POINT_2U other) => this == other;
+
+        public override bool Equals(object? obj) => (obj is D2D_POINT_2U other) && this == other;
+
+        public override int GetHashCode() => HashCode.Combine(x, y);
+
+        public static bool operator ==(D2D_POINT_2U left, D2D_POINT_2U right) => left.x == right.x && left.y == right.y;
```
tannergooding: nit: operators should be defined after constructors/properties, but before instance methods or other static methods.

sources/Interop/D2D1/um/d2d1/D2D1_BITMAP_BRUSH_PROPERTIES.Manual.cs
```diff
@@ -0,0 +1,19 @@
...
+// Ported from um/d2d1helper.h in the Windows SDK for Windows 10.0.18362.0
+// Original source is Copyright © Microsoft. All rights reserved.
+
+namespace TerraFX.Interop
+{
+    public partial struct D2D1_BITMAP_BRUSH_PROPERTIES
+    {
+        public static readonly D2D1_BITMAP_BRUSH_PROPERTIES DEFAULT = new D2D1_BITMAP_BRUSH_PROPERTIES(D2D1_EXTEND_MODE.D2D1_EXTEND_MODE_CLAMP, D2D1_EXTEND_MODE.D2D1_EXTEND_MODE_CLAMP, D2D1_BITMAP_INTERPOLATION_MODE.D2D1_BITMAP_INTERPOLATION_MODE_LINEAR);
+
+        public D2D1_BITMAP_BRUSH_PROPERTIES(D2D1_EXTEND_MODE extendModeX, D2D1_EXTEND_MODE extendModeY, D2D1_BITMAP_INTERPOLATION_MODE interpolationMode)
```
tannergooding: nit: We should probably specify the optional parameters here as well. Users can still choose to only configure a subset.

tannergooding: nit: The header file defines most of these as global helper functions rather than constructors.

I'm fine with them being exposed as constructors but we should still expose them as global helpers as well to make porting easier.

tannergooding:
Thanks!

PR #41, opened 2020-04-21
Fixed namespaces and class declarations

nathan-alden-sr:


sources/Interop/WinTrust/um/SoftPub/SoftPub.Manual.cs
```diff
@@ -1,8 +1,8 @@
 using System;
 
-namespace TerraFX.Interop.um.SoftPub
+namespace TerraFX.Interop
 {
-    public static class SoftPub
+    public static partial class SoftPub
```
tannergooding, line 5: This one should actually be named `WinTrust` since it is part of the `WinTrust` assembly. It helps simplify the general usage so you don't need to remember which include file it was part of.

nathan-alden-sr: Makes sense. Fixed.

PR #43, opened 2020-04-22
Mapped nearly all #defines in winerror.h

nathan-alden-sr:
I modified the conversion script to handle several new common macros. See the [Gist history](https://gist.github.com/nathan-alden-sr/63dea27baeed8dc47eb9fb4f3def7614/revisions) for changes.

### DIFF
--- a/sources/Interop/Windows/shared/winerror/Windows.Manual.cs
+++ b/sources/Interop/Windows/shared/winerror/Windows.Manual.cs
@@ -297,7 +297,7 @@ namespace TerraFX.Interop
 
         public const int NO_ERROR = 0;
 
-        public const int SEC_E_OK = 0x00000000;
+        public const int SEC_E_OK = unchecked((int)0x00000000);
 
         public const int ERROR_INVALID_FUNCTION = 1;
 
@@ -590,8 +590,6 @@ namespace TerraFX.Interop
         public const int ERROR_INVALID_EXE_SIGNATURE = 191;
 
         public const int ERROR_EXE_MARKED_INVALID = 192;
-
-        // %1 is not a valid Win32 application.;
 
         public const int ERROR_BAD_EXE_FORMAT = 193;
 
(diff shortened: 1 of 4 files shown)


sources/Interop/Windows/shared/winerror/Windows.Manual.cs
```diff
@@ -963,361 +961,12902 @@ public static partial class Windows
...
+
+        public const int ERROR_VRF_CFG_ENABLED = 1183;
+
+        public const int ERROR_PARTITION_TERMINATING = 1184;
+
+        public const int ERROR_USER_PROFILE_LOAD = 500;
+
         public const int ERROR_ARITHMETIC_OVERFLOW = 534;
 
-        public const int SEVERITY_SUCCESS = 0;
```
tannergooding, line 968: I wonder if we could do something to help minimize this diff...

This moved down to https://github.com/terrafx/terrafx.interop.windows/pull/43/files#diff-902801b79342b8c581d349536688d080R6432, so maybe if we inserted 5464 lines in a commit before the commit containing the new defines....

sources/Interop/Windows/shared/winerror/Windows.Manual.cs
```diff
@@ -963,361 +961,12902 @@ public static partial class Windows
...
+
+        public const int REGDB_E_CLASSNOTREG = unchecked((int)0x80040154);
+
+        public const int REGDB_E_IIDNOTREG = unchecked((int)0x80040155);
+
+        public const int REGDB_E_BADTHREADINGMODEL = unchecked((int)0x80040156);
+
+        public const int REGDB_E_PACKAGEPOLICYVIOLATION = unchecked((int)0x80040157);
+
+        public const long CAT_E_FIRST = 0x80040160L;
```
tannergooding, line 6746: `L` is C `long` which is not C# `long`. C `long` is `int` on Windows and `nint` on Unix.

tannergooding: `LL` is C `long long` which is C# `long`

PR #47, opened 2020-05-03
Generating bindings for hidclass.h and hidpi.h

nathan-alden-sr:
I decided to keep `hidusage.h` in the `Windows` project and also put `hidclass.h` there, since neither of those headers appear to define function exports. `hidpi.h` got its own project because it's dependent on `hid.dll`. I also migrated several macro functions and had to assume `int` as the type for them.


sources/Interop/Hid/shared/hidpi/HIDP_PREPARSED_DATA.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+// Ported from shared/hidpi.h in the Windows SDK for Windows 10.0.18362.0
+// Original source is Copyright © Microsoft. All rights reserved.
+
+namespace TerraFX.Interop
+{
+    public partial struct HIDP_PREPARSED_DATA
```
tannergooding, line 8: Looks like this is an opaque struct and is meant to be essentially a `handle` (`IntPtr`), I can only find a typedef such as `typedef struct _HIDP_PREPARSED_DATA * PHIDP_PREPARSED_DATA;` and no formal definition of the actual struct.

sources/Interop/Hid/shared/hidpi/Hid.Manual.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public static int HidP_GetButtons(HIDP_REPORT_TYPE ReportType, ushort UsagePage, ushort LinkCollection, ushort* UsageList, uint* UsageLength, HIDP_PREPARSED_DATA* PreparsedData, sbyte* Report, uint ReportLength) =>
+            HidP_GetUsages(ReportType, UsagePage, LinkCollection, UsageList, UsageLength, PreparsedData, Report, ReportLength);
+
+        public static int HidP_GetButtonsEx(HIDP_REPORT_TYPE ReportType, ushort LinkCollection, USAGE_AND_PAGE* ButtonList, uint* UsageLength, HIDP_PREPARSED_DATA* PreparsedData, sbyte* Report, uint ReportLength) =>
+            HidP_GetUsagesEx(ReportType, LinkCollection, ButtonList, UsageLength, PreparsedData, Report, ReportLength);
+
+        public static int HIDP_ERROR_CODES(int sev, int code) => ((sev) << 28) | (FACILITY_HID_ERROR_CODE << 16) | code;
+
+        public static readonly int HIDP_STATUS_SUCCESS = HIDP_ERROR_CODES(0x0, 0);
```
tannergooding, line 35: For these, the intent is that they are a `const`, similarly to how I did the errors in `winerror.h` even though they are defined using the various `MAKE_HRESULT` macros

sources/Interop/Windows/shared/hidclass/Windows.Manual.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public static int HID_CTL_CODE(int id) => CTL_CODE(FILE_DEVICE_KEYBOARD, id, METHOD_NEITHER, FILE_ANY_ACCESS);
+
+        public static int HID_BUFFER_CTL_CODE(int id) => CTL_CODE(FILE_DEVICE_KEYBOARD, id, METHOD_BUFFERED, FILE_ANY_ACCESS);
+
+        public static int HID_IN_CTL_CODE(int id) => CTL_CODE(FILE_DEVICE_KEYBOARD, id, METHOD_IN_DIRECT, FILE_ANY_ACCESS);
+
+        public static int HID_OUT_CTL_CODE(int id) => CTL_CODE(FILE_DEVICE_KEYBOARD, id, METHOD_OUT_DIRECT, FILE_ANY_ACCESS);
+
+        public static readonly int IOCTL_HID_GET_DRIVER_CONFIG = HID_BUFFER_CTL_CODE(100);
```
tannergooding, line 30: Same here. The intent behind the macros is that they are `const`, if they were modern C++ they would likely be `constexpr` methods instead.

sources/Interop/Hid/shared/hidpi/HIDP_PREPARSED_DATA.cs
nathan-alden-sr: From what I read in a Code Project sample I've been using, the data is internal and only intended to be used by the system.
> The Raw Input API gives us access to some device information using the GetRawInputDeviceInfo function. Given the handle to the raw input device and the RIDI_PREPARSEDDATA command we can obtain so called "preparsed data". The WDK documentation just says, "The internal structure of a _HIDP_PREPARSED_DATA structure is reserved for internal system use".
https://www.codeproject.com/articles/185522/using-the-raw-input-api-to-process-joystick-input

I know this might not be authoritative but it's all I have to go on. Let me know how you'd like to handle this one.

nathan-alden-sr: I would favor keeping it since it's a type logically, even if not physically. From my exploration it seems the way one is supposed to allocate space for this type is to allocate a byte array and then cast the type. It's that oddball.

sources/Interop/Hid/shared/hidpi/Hid.Manual.cs
nathan-alden-sr: Are you referring to these?
```cs
public static int MAKE_D2DHR(int sev, int code) => MAKE_HRESULT(sev, FACILITY_D2D, code);
```
I see the only difference is field vs. property. Does the `=>` lead to the compiler treating this as a constant somehow? I'm not sure what it is I should fix.

sources/Interop/Hid/shared/hidpi/HIDP_PREPARSED_DATA.cs
nathan-alden-sr: I guess by "opaque struct" you're maybe identifying a concept I'm not that familiar with. Maybe this pattern isn't so oddball. Let me know if you'd like me to map it to `IntPtr` and I will.

tannergooding: Right, it's meant to be completely "opaque". There are similar in other libraries (`XDisplay` on `X11`) and the general point is to the average user it is just a handle, but to a sufficiently advanced user you could try and access the underlying bits (you can do the same with `HWND` which, depending on your defines is either a typedef for a `void*` or a typedef to an opaque `struct __HWND`).

The general convention for these is to make it an `IntPtr` and the dev can define and cast the pointer to their own type if they really want to access internals.

tannergooding: By "opaque" I mean it is a struct that completely hides the internals from the user (in this case, it exposes nothing in the header and you only have a pointer to it, which you can pass to other methods) vs a "transparent" one where you can see, access, and depend on the members directly.

nathan-alden-sr: https://en.wikipedia.org/wiki/Opaque_pointer was a good read on the topic. Thanks for pointing this out.

sources/Interop/Hid/shared/hidpi/Hid.Manual.cs
tannergooding: No, the member would have to be redeclared as a `public const int HIDP_STATUS_SUCCESS = resultOfMacro` rather than being a `static readonly` that calls a helper method.

Initializing all these statics has a cost and prevents some downstream optimizations and the C/C++ compile treats these type of macros as actual constants as well.
It's probably not something that is easily handled in the Powershell script and I've done it manually up to this point (as I haven't finished getting macro support added to ClangSharp).

nathan-alden-sr: There aren't too many. I can do it manually for now.

Issue #46, opened 2020-05-03
[IDEA] .Net 5.0 + Dx12 Hello World


IngmarBitter:
TerraFx seems to be the ideal project for demonstrating best practices using modern Microsoft solutions for modern apps.
I see as a personal immediate need, that probably is actually quite common, to have to write a Windows 10 app that should have a fluid GUI (.NET 5.0 with WinUI 3.0 would be ideal) and some panels/viewports in the window with high performance 3D graphics (D3D12 would be ideal) but the graphics viewports are not a game but a technical interface and thus also need a toolstrip on the side (ideally semitransparent) and a mouse right click context menu.
It seems that TerraFx already has most of these technologies combined and would have expanded appeal if there was a Hello World that would demonstrate how to use them together.

A minimal desirable set of functionality would be:
1) A main window using XAML that can be edited in the Visual Studio Designer. On the top level a grid with two columns:
   Left a button from WinUI 3.0
   Right a panel with a D3D12 display viewport in it.
2) Upon start the D3D12 display shows the triangle from the current D3D12 hello triangle demo.
3) Upon button press the triangle changes color
4) On the D3D12 display is a toolstrip with button (with an icon being a triangle over a transparent background) that upon press also changes the triangle color.
5) Right click anywhere on the D3D12 display brings up a context menu with one entry to change color.
6) Left mouse drag on the D3D12 display rotates or zooms the triangle.
7) Be a model of best practices, so include the fence demo setup from D3D12HelloFrameBuffering
8) A second button that triggers creating a simple checkerboard image in C# and then sending it to D3D12 as texture for the triangle as an example how to move memory from the CPU to the GPU.

tannergooding:
I'm fine with providing additional samples that show how to correctly integrate the interop bindings with various UI Frameworks (WinForms, WPF, UWP, WinUI, etc), but I expect most of that to come after I'm able to rewrite the bindings to use function pointers and move closer to a v1.0 release of the bindings proper.

IngmarBitter:
I can probably help with making some of the samples (WinForms & WinUI) for Dx12. Might just need a little guidance, but it will likely still save you time. We could review progress via Discord or MS Teams.

tannergooding:
The invite to the TerraFX discord is: https://discord.terrafx.dev/

PR #52, opened 2020-05-05
WinForms sample project: hello triangle

IngmarBitter:
[IMPL]Proposal Implementation

WinFormsDx12 folder and project added.
WinForms app is running and using TerraFX to render a triangle that shows upon load and background color changes upon pressing [Color] button or resize.
Dx12Panel class added that takes care of Dx12 context creation and calling its render method upon Paint and Resize
DXSample.OnResize() added as start of response to Resize event (work in progress as the Dx12 pipeline state needs to be updated to reflect the new size).
refs #46


samples/WinFormsDx12/Program.cs
```diff
@@ -0,0 +1,23 @@
+using System;
```
tannergooding, line 1: All of the files need a header similar to `// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.`

samples/WinFormsDx12/WinFormsDx12.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
```
tannergooding, line 1: Same here, but the header should be:
```
<?xml version="1.0" encoding="utf-8"?>
<!-- Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information. -->
```

IngmarBitter: added copyright notices to headers and a little more code cleanup.

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
```
tannergooding, line 1:
```suggestion
// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.

```

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using System.Windows.Forms;
+using TerraFX.Samples.DirectX.D3D12;
+
+
```
tannergooding, line 6:
```suggestion

```

samples/GuiSamples.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
tannergooding, line 1: This file could probably be named `TerraFX.Interop.Windows.Samples.sln` and be in the root next to `TerraFX.Interop.Windows.sln` for discoverability

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace WinFormsDx12
```
tannergooding, line 6:
```suggestion
namespace TerraFX.Samples.WinForms
```

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public class Dx12Panel : Panel
```
tannergooding, line 8:
```suggestion
    public class DX12Panel : Panel
```

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
...
+        private readonly Dx12Viewport _dxVp;
+        public Dx12Viewport DxVp => _dxVp;
+
+        public Dx12Panel(int width, int height) : base()
+        {
+            _dxVp = InitDxVp(width, height);
+            Resize += Dx12Panel_Resize;
+        }
+
+        private Dx12Viewport InitDxVp(int width, int height)
```
tannergooding, line 19:
```suggestion
        private DX12Viewport InitializeViewport(int width, int height)
```

samples/WinFormsDx12/Dx12Panel.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var dxVp = new Dx12Viewport((uint)width, (uint)height, "Dx12Vp", Handle);
+            dxVp.OnInit();
+            return dxVp;
+        }
+
+        private void Dx12Panel_Resize(object? sender, EventArgs e)
```
tannergooding, line 26:
```suggestion
        private void OnResize(object? sender, EventArgs eventArgs)
```

samples/WinFormsDx12/Dx12Viewport.cs
```diff
@@ -0,0 +1,684 @@
...
+using static TerraFX.Interop.DXGI;
+using static TerraFX.Interop.DXGI_FORMAT;
+using static TerraFX.Interop.DXGI_SWAP_EFFECT;
+using static TerraFX.Interop.Kernel32;
+using static TerraFX.Interop.Windows;
+using static TerraFX.Samples.DirectX.DXSampleHelper;
+
+namespace TerraFX.Samples.DirectX.D3D12
+{
+    public unsafe class Dx12Viewport : DX12Sample
```
tannergooding, line 46:
```suggestion
    public unsafe class DX12Viewport
```

Noting that using `DX12Sample` or `DXSample` may be problematic, since it uses `Win32Application` for some bits and that won't work with `WinForms`

samples/WinFormsDx12/Dx12Viewport.cs
```diff
@@ -0,0 +1,684 @@
...
+
+            var rtvHandle = _rtvHeap->GetCPUDescriptorHandleForHeapStart();
+            rtvHandle.ptr = (UIntPtr)((byte*)rtvHandle.ptr + (_frameIndex * _rtvDescriptorSize));
+            _commandList->OMSetRenderTargets(1, &rtvHandle, FALSE, null);
+
+            // Record commands.
+            var r = _random.Next(0, 50) / 100.0f;
+            var g = _random.Next(0, 50) / 100.0f;
+            var b = _random.Next(0, 50) / 100.0f;
+            var clearColor = stackalloc float[4] { 0.0f, r, g, b };
```
tannergooding, line 541: It would be better to either toggle between two well-defined colors or use a color picker control and letting the user select the target color

samples/WinFormsDx12/MainFormWinFormsDx12.cs
```diff
@@ -0,0 +1,35 @@
...
+            return dxPanel.DxVp;
+        }
+
+        private void ColorBtn_Click(object sender, EventArgs e) => _dxVp.OnRender();
+
+        private void TextureBtn_Click(object sender, EventArgs e)
+        {
+
+        }
+
```
tannergooding, line 33:
```suggestion
```

samples/WinFormsDx12/Dx12Viewport.cs
IngmarBitter: Implemented toggle between two colors. The idea is just to be able to see that it indeed re-rendered the scene in response to the button or resize event.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+using System;
+using System.Windows.Forms;
+using TerraFX.Samples.DirectX.D3D12;
+namespace TerraFX.Samples.WinForms
+{
+    public class DX12Panel : Panel
+    {
+        private readonly DX12Viewport _dxVp;
```
tannergooding, line 10: It would be nice to just name the field `_viewport`, it will help with readability of the code

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public DX12Panel(int width, int height) : base()
+        {
+            _dxVp = InitializeViewport(width, height);
+            Resize += OnResize;
+        }
+
+        private DX12Viewport InitializeViewport(int width, int height)
+        {
+            var dxVp = new DX12Viewport((uint)width, (uint)height, Handle);
```
tannergooding, line 21: `width` and `height` should be validated to be non-negative if this is being done, otherwise users will pass in a value and it may wrap around.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,40 @@
...
+            return dxVp;
+        }
+
+        private void OnResize(object? sender, EventArgs eventArgs)
+        {
+            _dxVp.OnResize((uint)Width, (uint)Height);
+            // _dxVp.OnInit(); // IB: crashes on resoure reinitialization, needs instead a _dxVp.OnResize that updates existing resources
+            _dxVp.OnRender();
+        }
+
```
tannergooding, line 32: There is a newline here that needs to be removed
```suggestion
```

samples/WinForms/Dx12Viewport.cs
```diff
@@ -0,0 +1,769 @@
...
+using static TerraFX.Interop.DXGI_SWAP_EFFECT;
+using static TerraFX.Interop.Kernel32;
+using static TerraFX.Interop.Windows;
+using static TerraFX.Samples.DirectX.DXSampleHelper;
+
+namespace TerraFX.Samples.DirectX.D3D12
+{
+    public unsafe class DX12Viewport : IDisposable
+    {
+
```
tannergooding, line 38: There is a newline here that needs to be removed
```suggestion
```

samples/WinForms/Dx12Viewport.cs
```diff
@@ -0,0 +1,769 @@
...
+        private readonly ID3D12Resource*[] _renderTargets;
+        private ID3D12CommandAllocator* _commandAllocator;
+        private ID3D12CommandQueue* _commandQueue;
+        private ID3D12RootSignature* _rootSignature;
+        private ID3D12DescriptorHeap* _rtvHeap;
+        private ID3D12PipelineState* _pipelineState;
+        private ID3D12GraphicsCommandList* _commandList;
+        private uint _rtvDescriptorSize;
+        private readonly IntPtr _hWnd;
+        private bool _isOnColorToggle;
```
tannergooding, line 68: This really shouldn't be part of the viewport itself... Ideally the viewport would just be a generic/reusable and sample specific things would be tracked in the respective sample.

I'm also not sure `Viewport` is a good name here, this is really all the bits needed for the `DX12Panel` and the two classes should probably be merged.
Bits like the `Width`, `Height`, and `_hWnd` would then come from the `Panel` object that `DX12Panel` inherits from.

samples/WinForms/Dx12Viewport.cs
```diff
@@ -0,0 +1,769 @@
...
+        // App resources.
+        private ID3D12Resource* _vertexBuffer;
+        private D3D12_VERTEX_BUFFER_VIEW _vertexBufferView;
+
+        // Synchronization objects.
+        private uint _frameIndex;
+        private HANDLE _fenceEvent;
+        private ID3D12Fence* _fence;
+        private ulong _fenceValue;
+
```
tannergooding, line 79: There is a newline here that needs to be removed
```suggestion
```

samples/WinForms/Dx12Viewport.cs
```diff
@@ -0,0 +1,769 @@
...
+        // Samples override the event handlers to handle specific messages
+        public virtual void OnKeyDown(byte key)
+        {
+        }
+
+        public virtual void OnKeyUp(byte key)
+        {
+        }
+
```
tannergooding, line 135: There is a newline here that needs to be removed
```suggestion
```

samples/WinForms/DXSampleHelper.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
```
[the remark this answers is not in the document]
IngmarBitter: good suggestion. Will do.

samples/WinForms/Dx12Panel.cs
IngmarBitter: agreed.

samples/WinForms/Dx12Panel.cs
IngmarBitter: assuming this is an appropriate check and response:
            `if (width <= 0 || height <= 0) throw new ArgumentException("width and height must be positive and non-zero");`
Let me know if you had something else in mind.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            var dxVp = new DX12Viewport((uint)width, (uint)height, Handle);
+            dxVp.OnInit();
+            return dxVp;
+        }
+
+        private void OnResize(object? sender, EventArgs eventArgs)
+        {
+            _dxVp.OnResize((uint)Width, (uint)Height);
+            // _dxVp.OnInit(); // IB: crashes on resoure reinitialization, needs instead a _dxVp.OnResize that updates existing resources
```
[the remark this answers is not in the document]
IngmarBitter: // IB is what I use as my initials to mark comments that I need to get back to and fix.  The Resizing is supposed to create a new rasterization of the triangle at the size matching the new window size. But currently it just scales the prior image. Needs a little more DX12 understanding than I have so far. But I'll get there soon.  The current state of this project is not final, but a first draft so I get a feel for what programming style you are looking for.

samples/WinForms/Dx12Viewport.cs
IngmarBitter: done.

samples/WinForms/Dx12Viewport.cs
IngmarBitter: ok

samples/WinForms/Dx12Viewport.cs
IngmarBitter: ok

samples/WinForms/Dx12Panel.cs
IngmarBitter: ok

samples/WinForms/Dx12Viewport.cs
IngmarBitter: Makes sense.  Can I do that in a subsequent PR?

samples/WinForms/Dx12Panel.cs
tannergooding: If https://github.com/terrafx/terrafx.interop.windows/pull/52#discussion_r421548758 is taken care of, then `Panel` should end up validating these parameters for you.

samples/WinFormsDx12/MainFormWinFormsDx12.cs
IngmarBitter: done

samples/WinForms/Dx12Viewport.cs
IngmarBitter: actually, ignore the above comment. Refactoring now. Will be a separate commit. That should be enough to encapsulate this change. Does not need a new PR.

IngmarBitter: done.

IngmarBitter:
update on implementation state:
Project is now called TerraFX.Samples.WinForms.
It is a WinForms GUI with a Panel that contains DirectX content.
The DirectX content is taken from the Hello Triangle example.
The main point of the project is to demonstrate how to connect a WinForms GUI to DirectX rendering.

samples/WinForms/Assets/Shaders/HelloTriangle.hlsl
```diff
@@ -0,0 +1,25 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
```
tannergooding, line 1: This is identical to the existing shader from the more general samples, it should likely be linked rather than copied.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,868 @@
...
+using static TerraFX.Interop.Kernel32;
+using static TerraFX.Interop.Windows;
+using static TerraFX.Samples.DirectX.DXSampleHelper;
+
+
+namespace TerraFX.Samples.WinForms
+{
+    public unsafe class DX12Panel : Panel
+    {
+        #region local structs
```
tannergooding, line 40: We shouldn't be adding regions to the code.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,868 @@
...
+
+        private struct Vertex
```
tannergooding, line 42: Structs should appear at the bottom of the type. The ordering is basically `fields`, `constructors`, `destructors`, `delegates`, `events`, `enums`, `interfaces`, `properties`, `indexers`, `methods`, `structs`, `classes`

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,868 @@
...
+        private ID3D12Device* _device;
+        private readonly ID3D12Resource*[] _renderTargets;
+        private ID3D12CommandAllocator* _commandAllocator;
+        private ID3D12CommandQueue* _commandQueue;
+        private ID3D12RootSignature* _rootSignature;
+        private ID3D12DescriptorHeap* _rtvHeap;
+        private ID3D12PipelineState* _pipelineState;
+        private ID3D12GraphicsCommandList* _commandList;
+        private uint _rtvDescriptorSize;
+        private bool _isOnColorToggle;
```
tannergooding, line 76: This isn't relevant to the `Panel` and is instead a feature of the sample, it shouldn't be here.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,868 @@
...
+
+        public bool UseWarpDevice => _useWarpDevice;
+
+        private Vector2 _mousePriorPt;
+
+        #endregion data members
+
+        #region constructor
+
+        public DX12Panel(int width, int height, bool useWarpDevice = false) : base()
```
tannergooding, line 98: Note to self: There is a lot of duplication between this type and the underlying non-WinForms samples. Ideally we could have a shared abstraction that `DX12Panel` just wraps and we could support rendering any of the samples in WinForms with no changes required.

This would also simplify things when the sample is extended to WPF or other UI frameworks.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,868 @@
...
+            MouseMove += OnMouseMove;
+            MouseWheel += OnMouseWheel;
+            OnInit();
+        }
+
+        #endregion constructor
+
+        #region mouse event handling
+
+        private void OnMouseEnter(object? sender, EventArgs e)
```
tannergooding, line 148: We should be overriding the existing virtual members we inherited from `Panel`.

tannergooding: I'm also not sure these should be on here. There isn't really any specific functionality that we should be overriding and instead each sample should be hooking into the `DX12Panel` to change what is rendered, etc.

samples/WinForms/MainFormWinFormsDx12.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private readonly DX12Panel _panel;
+
+        public MainFormWinFormsDX12()
+        {
+            InitializeComponent();
+            _panel = InitDX12Panel();
+        }
+
+        private DX12Panel InitDX12Panel()
```
tannergooding, line 25: Shouldn't this all be done in the designer?

samples/WinForms/MainFormWinFormsDx12.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var dxPanel = new DX12Panel(MainSplitter.Panel2.Width, MainSplitter.Panel2.Height);
+            MainSplitter.Panel2.Controls.Add(dxPanel);
+            dxPanel.Dock = DockStyle.Fill;
+            return dxPanel;
+        }
+
+        private void ColorBtn_Click(object sender, EventArgs e) => _panel.OnRender();
```
tannergooding, line 33: I don't think this is layered correctly. The panel should self render at the appropriate frame rate, not on demand.

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: The designer does not let me use the custom panel :(

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: No its correct. While there is no change to the graphics scene there should be no re-render. Compute resources should be spent only when needed.

samples/WinForms/Assets/Shaders/HelloTriangle.hlsl
IngmarBitter: ok

samples/WinForms/MainFormWinFormsDx12.cs
tannergooding: What is the error it gives?

samples/WinForms/MainFormWinFormsDx12.cs
tannergooding: That's not how multimedia panels typically work. By default, you render at a given FPS and you only fallback to on demand if requested. Otherwise animations and the like don't work.

IngmarBitter: This is especially true for the application I have in mind at the end. There the rendering will be so expensive that it can not be done at an appropriate frame rate. Hence I'll have to simplify the scene while the user interacts with it and render the full quality when interaction is over (upon mouse up event).
This has no equivalent in a game. There everything must be same quality all the time.

IngmarBitter: hmm, I can add a refresh timer to render an a particular frame rate. But then something has to change for it to be visual.   Let's keep this PR as is. Then I'll make another to port the HelloConstantBuffer to TerraFX.Interop.Windows.sln. When that works I'll reapply it to the WinForms project with a timer that calls OnRender at 60Hz. But I'd really like that to come from a new branch PR, not this one that is on the master.

tannergooding: Sure, but this sample isn't for your game, its for the general audience coming to TerraFX.Interop.Windows.

I'm fine with the sample panel having a built in mechanism to specify the target framerate and even for that to be `0` so that you can make it explicit. But the default should likely just be the same as the rest of the samples.

samples/WinForms/Dx12Panel.cs
IngmarBitter: I don't get what you are suggesting.  The use case I had in mind was that I drag rotate the triangle with the mouse. For that I need to implement OnMouseMove() and if I want the cursor to change to a cursor that indicates rotation then I need to set that in OnMouseEnter() and revert it in OnMouseLeave().  That is what these placeholders are there for.  How would you handle the cursor change and mouse drag rotation differently?

samples/WinForms/Dx12Panel.cs
IngmarBitter: Agreed.  I think it stems from the original C++ examples that have a lot of shared code, too. They seem to have been made via copy and modify.  But now that they are that way its probably best to have the C# version be as similar as possible. So within TerraFX.Interop.Windows.sln the duplication is fine.  But still in TerraFX.Interop.Windows.Samples.sln where there is no C++ reference reuse would be better.  Something to refactor when we have the second one?

samples/WinForms/Dx12Panel.cs
IngmarBitter: I'm fine with removing them.  I thought they make the code more readable. Why do you not like them?

samples/WinForms/Dx12Panel.cs
IngmarBitter: ok, will do.

samples/WinForms/Dx12Panel.cs
tannergooding: `Panel` already defines `OnMouseMove` (and the others), so they should be overridden rather than redeclared. But, these also shouldn't be overridden until they are actually being used.

samples/WinForms/Dx12Panel.cs
tannergooding: Most code bases don't use regions as they typically indicate a class is doing too much and would be better broken up into better abstractions or split into multiple files.

Consistent ordering helps deal with readability and ease of locating particular members.

samples/WinForms/Dx12Panel.cs
IngmarBitter: hmm, so all the scene description and rendering should go to MainFormWinFormsDX12? Or better yet to a SceneModel class that then can also be used by a WPF sample and others?  Sounds good. Can that be done on a different PR (see next comment also)?

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: It needed to be a separate project and I think it did not like the .Net 5.0 preview. But I need to find the instructions again on making a custom panel ...

samples/WinForms/Assets/Shaders/HelloTriangle.hlsl
IngmarBitter: done

samples/WinForms/Dx12Panel.cs
IngmarBitter: done

samples/WinForms/Dx12Panel.cs
IngmarBitter: OK. That is fine. Then I'll remove them for now.

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: I was following https://docs.microsoft.com/en-us/dotnet/framework/winforms/controls/creating-a-wf-control-design-time-features
Will try again.

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: ok

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: So what got me off the track of making a designer enabled panel was that it requires a new project and even Visual Studio 2019 Preview has no option for WinForms control library plus .Net Core or .Net 5.0. So the project file style and the target framework both do not match.  So I stayed with adding the panel at runtime.

samples/WinForms/Dx12Panel.cs
IngmarBitter: I removed the regions.

samples/WinForms/MainFormWinFormsDx12.cs
tannergooding: It shouldn't require a new project, but it does look like the designer for .NET Core doesn't support custom controls yet and you'll get an error if you try to drag/drop.

samples/WinForms/Dx12Panel.cs
tannergooding: I'm fine with it being handled in a follow up PR. Please log an issue to make sure it doesn't get forgotten however.

samples/WinForms/Dx12Panel.cs
tannergooding: I was actually thinking of just tweaking the class that holds the sample slightly so it could be reused here as well. The general port would still match but each sample, with minimal modification, could be tweaked so it can be hosted anywhere.

I'll likely prototype something here myself.

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,761 @@
...
+using static TerraFX.Interop.D3DCompiler;
+using static TerraFX.Interop.DXGI;
+using static TerraFX.Interop.DXGI_ADAPTER_FLAG;
+using static TerraFX.Interop.DXGI_FORMAT;
+using static TerraFX.Interop.DXGI_SWAP_EFFECT;
+using static TerraFX.Interop.Kernel32;
+using static TerraFX.Interop.Windows;
+using static TerraFX.Samples.DirectX.DXSampleHelper;
+
+
```
tannergooding, line 35: One too many newlines here
```suggestion
```

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,761 @@
...
+                }
+
+                if (pixelShader != null)
+                {
+                    pixelShader->Release();
+                }
+            }
+        }
+
+        private void OnResize(object? sender, EventArgs e)
```
tannergooding, line 529: This should override `OnSizeChanged` (or maybe `OnClientSizeChanged`, since that is the area that we really care about).

samples/WinForms/Dx12Panel.cs
```diff
@@ -0,0 +1,761 @@
...
+        {
+            _aspectRatio = Width / ((float)Height);
+            OnRender();
+        }
+
+        protected override void OnPaint(PaintEventArgs e)
+        {
+            base.OnPaint(e);
```
tannergooding, line 537: I don't think we need to call `base.OnPaint(e)`

tannergooding:
This should be ready to merge after the issue gets filed and the couple remaining asks get fixed.

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: Right. So for now it has to be added at runtime. Which means not great, but nothing to change for now.

samples/WinForms/Dx12Panel.cs
IngmarBitter: created issue https://github.com/terrafx/terrafx.interop.windows/issues/61

samples/WinForms/Dx12Panel.cs
IngmarBitter: Sure, please tweak it.  
Also see https://github.com/terrafx/terrafx.interop.windows/issues/61 that addresses the other similar concern you had.  I suppose for now that means no further change on this PR due to this particular problem.

samples/WinForms/MainFormWinFormsDx12.cs
IngmarBitter: Logged issue https://github.com/terrafx/terrafx.interop.windows/issues/62 so we do not forget about this.  I'll implement after HelloConstantBuffer  is working. So no further change needed under this PR.

samples/WinForms/Dx12Panel.cs
IngmarBitter: I switched it to overriding OnClientSizeChanged(). It triggers calling OnPaint automatically, so no extra call to that needed. Seems indeed better using this inheritance mechanism than adding an event handler.

samples/WinForms/Dx12Panel.cs
IngmarBitter: Correct. Works fine without. Fixed.

PR #53, opened 2020-05-05
Fix compiler messages in sample projects

IngmarBitter:
[Bug Fix] Sorry, no bug#. Feel free to ignore :)
This fix creates a clean build of the samples solution with all messages of the compiler attended to.
Biggest change is replacing
`protected type _name;`
with
`private type _name;  public type Name => _name;`

### DIFF
--- a/samples/DirectX/D3D12/HelloTriangle12.cs
+++ b/samples/DirectX/D3D12/HelloTriangle12.cs
@@ -52,7 +52,7 @@ namespace TerraFX.Samples.DirectX.D3D12
         private RECT _scissorRect;
         private IDXGISwapChain3* _swapChain;
         private ID3D12Device* _device;
-        private ID3D12Resource*[] _renderTargets;
+        private readonly ID3D12Resource*[] _renderTargets;
         private ID3D12CommandAllocator* _commandAllocator;
         private ID3D12CommandQueue* _commandQueue;
         private ID3D12RootSignature* _rootSignature;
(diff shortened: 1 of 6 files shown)


samples/DirectX/D3D12/HelloTriangle12.cs
```diff
@@ -111,11 +111,11 @@ public override void OnRender()
             PopulateCommandList();
 
             // Execute the command list.
-            const int ppCommandListsCount = 1;
-            var ppCommandLists = stackalloc ID3D12CommandList*[ppCommandListsCount] {
+            const int PPCommandListsCount = 1;
```
tannergooding, line 114: Just calling this `CommandListCount` would likely be better.

samples/DirectX/Shared/DXSample.cs
```diff
@@ -48,6 +48,11 @@ protected DXSample(uint width, uint height, string name)
         public uint Width => _width;
 
         public uint Height => _height;
+        public float AspectRatio => _aspectRatio;
```
tannergooding, line 51: There is a missing newline between this member and the previous

samples/GuiSamples.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
tannergooding, line 2: These files shouldn't be part of this PR, they are part of https://github.com/terrafx/terrafx.interop.windows/pull/52

tannergooding:
It would be good to clean up the commits rather than reverting those from the other PR. The repo does normal merging and this unnecessarily pollutes the history.

IngmarBitter:
> It would be good to clean up the commits rather than reverting those from the other PR. The repo does normal merging and this unnecessarily pollutes the history.

Do you mean collapse commits so that the file adds and removes do not show up? I agree in principle. The problem is that I created the branch after I had make changes on the local master. So I see no option for collapsing commits in the git history. Best to just delete the branch and restart on a proper clean branch?

tannergooding:
You have a couple options such as `cherry-pick` or `rebase`.

The former is likely simpler and would be (assuming you or your tooling set `upstream` to `terrafx/terrafx.interop.windows` when you cloned the repo):
```
git fetch --all
git reset --hard upstream/master
git cherry-pick ac5c6ede5b1b54175553d064a766a1f195223fc3
git cherry-pick 3584eb4aa8670be85c06866d8ca809c89a508e65
```

The latter requires you to have either configured an alternative editor or to be comfortable with vim. You would do:
```
git rebase -i HEAD~8
```

This would pop up a text window showing the last 8 commits and some explanatory text telling you that you can change `pick` to `drop` to no longer include the commit in the rewritten history. Saving and closing that text causes the selected actions to be run.
Given that you have a merge commit, it might be slightly more complicated than that and is more advanced to do in general.

tannergooding:
Generally, you don't want to close and open a new PR here since it just creates noise on the repo you are contributing to and instead you want to fixup the existing PR (that may not always be the case, but in this simple scenario, that is likely best)

PR #54, opened 2020-05-06
Manual IO reparse macro mappings

nathan-alden-sr:
I did these by hand since `winnt.h` is 22000+ lines. I'll get to the whole header eventually.


nathan-alden-sr:
Don't review yet; I found more reparse tags.

nathan-alden-sr:
[This](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4) document defines 54 reparse tags but `winnt.h` only defines 45 of them. I omitted the ones not defined in `winnt.h` since the header is the canonical source.

PR #58, opened 2020-05-11
Remove unnecessary event object

john-h-k:
I think this is clearer for beginners new to DX, who are probably a target of these samples. Semantically identical


tannergooding:
This is "simpler", but it isn't how the native sample runs and sets a foundation that you really need later even if you aren't using it quite yet.

Issue #48, opened 2020-05-04
[PROPOSAL] Update the contributing docs to include minimum tooling requirements.


tannergooding:
#### Rationale

Developers looking to contribute may not always use the command line or stay up to date with the latest tooling/previews and so they may run into accidental issues when the repo moves forward.
As such, it would be beneficial if the contributing docs and potentially the readme clearly called out the tooling requirements for the repository.
This includes both the command line and graphical experience via an IDE like Visual Studio.

#### Alternatives

Some of this could be simplified by enabling experimental work as opt-in or by doing it in a separate branch. However, developers would still want to know those dependencies and so something like the above may still be required.

PR #64, opened 2020-05-17
[IMPL] Issue #48 contributing docs with additional helpful details

IngmarBitter:
[IMPL] #48 updated Readme.md and Contributing.md to better document what is needed to get started.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -7,9 +7,9 @@ The following is a set of guidelines for contributing to TerraFX.
 * [Code of Conduct](#code-of-conduct)
 * [License](#license)
 * [What should I know?](#what-should-i-know)
-* [Pull Requests](#pull-requests)
 * [Issues](#issues)
 * [Questions](#questions)
+* [Pull Requests](#pull-requests)
 
 ### Code of Conduct
 
@@ -28,6 +28,16 @@ Copyright © Tanner Gooding and Contributors. Licensed under the MIT License
 Most of the basics of the project, such as what it is about, and its goals are
 covered in our [README](README.md).
 
+### Issues
+
+All issues should follow our [ISSUE_TEMPLATE](ISSUE_TEMPLATE.md). It is
+additionaly recommended to prefix your issue with an appriopriate "tag" such as
+`[QUESTION]`, `[BUG]`, `[REGRESSION]`, `[IDEA]`, or `[PROPOSAL]`.
+
+### Questions
+
+Feel free to open an issue prefixed with `[QUESTION]`.
+
 ### Pull Requests
 
 All pull requests should follow our
@@ -38,12 +48,3 @@ template will be requested to be updated. Any pull requests opened without a
 corresponding issue may be delayed or be required to undergo further changes
 before being accepted.
 
-### Issues
-
-All issues should follow our [ISSUE_TEMPLATE](ISSUE_TEMPLATE.md). It is
-additionaly recommended to prefix your issue with an appriopriate "tag" such as
-`[QUESTION]`, `[BUG]`, `[REGRESSION]`, `[IDEA]`, or `[PROPOSAL]`.
-
-### Questions
-
-Feel free to open an issue prefixed with `[QUESTION]`.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -48,3 +48,79 @@ template will be requested to be updated. Any pull requests opened without a
 corresponding issue may be delayed or be required to undergo further changes
 before being accepted.
 
+#### Pull Request Workflow
+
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -38,12 +48,79 @@ template will be requested to be updated. Any pull requests opened without a
 corresponding issue may be delayed or be required to undergo further changes
 before being accepted.
 
-### Issues
+#### Pull Request Workflow
```
tannergooding, line 51: I don't think this belongs in the repo.

If anything, there should just be a simple link to the existing GitHub docs such as: 
* https://help.github.com/en/github/collaborating-with-issues-and-pull-requests

docs/CONTRIBUTING.md
```diff
@@ -38,12 +48,79 @@ template will be requested to be updated. Any pull requests opened without a
...
+a review you can make another branch for another PR and work on that in parallel.
+
+7. When a PR is approved and closed your code becomes part of the upstream code base 
+and there is no need to keep the branch. 
+Feel free to delete the branch to reduce clutter in your git log, 
+but this step is not required.
+
+#### Coding Conventions
+
+TerraFX follows the industry standard
```
tannergooding, line 82: This isn't quite true. I deviate from the Visual Studio defaults on a few places and there isn't necessarily an "industry standard" but rather recommended practices.

docs/CONTRIBUTING.md
```diff
@@ -38,12 +48,79 @@ template will be requested to be updated. Any pull requests opened without a
...
+TerraFX follows the industry standard 
+[C# coding conventions as published by Microsoft]
+(https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions) 
+and as enforced by the messages in Visual Studio 2019 
+when configured as *build + intellisense* .  
+Commits should be done such that under these settings there are no warnings.
+
+In addition, there are a few more conventions:
+
+1. **Property vs Method:** a property is auto evaluated by Visual Studio
```
tannergooding, line 91: While this is generally good advice, it doesn't belong here. `TerraFX.Interop.*` are all generally interop bindings and so they break almost all of the standard coding conventions/patterns.

Instead, rules like this should belong in a more general doc describing how devs submitting API proposals should consider the shape of the API before logging an issue.

docs/CONTRIBUTING.md
```diff
@@ -38,12 +48,79 @@ template will be requested to be updated. Any pull requests opened without a
...
+Commits should be done such that under these settings there are no warnings.
+
+In addition, there are a few more conventions:
+
+1. **Property vs Method:** a property is auto evaluated by Visual Studio 
+when stopping at a breakpoint for items in the watch/autos/locals panel. 
+If that takes long Visual Studio becomes laggy. 
+So if your property would take long to evaluate then better make it a method.
+
+2. **Compound Names:** Auto completion is very useful to use unfamiliar APIs.
```
tannergooding, line 96: 2 and 3 aren't conventions  that I am currently following.

For 2, it entirely depends on the API and in either case it is expected that IntelliSense will largely help as if you type `Distance` the autocomplete list will also show `MaxDistances` as an available option.
For 3, two letter postfixes goes against the general naming guidelines.

docs/CONTRIBUTING.md
IngmarBitter: The link leads to something a lot more complicated than what I wrote. But it is indeed not what the repo is about. So I'll replace the description with the link.

docs/CONTRIBUTING.md
IngmarBitter: But it is true that the .editor config sets the coding style to what you like? So if I resolve all warnings then it should be correct?

tannergooding: The `.editorconfig` specifies the repo conventions yes; but there are a few issues with it that cause it to deviate.

In some cases there are also other `.editorconfig` files which specify different rules. For example, with the actual interop bindings they are meant to match the native conventions, so you can't really define anything that "always matches"

docs/CONTRIBUTING.md
IngmarBitter: Well, they are things that I try to go by as much as possible. Applies more to the samples than to the core of TerraFX. So I'll remove it. Was there in part to get your feedback on it. So it will be part of my PRs, but not of this doc.

docs/CONTRIBUTING.md
IngmarBitter: Ok, will remove as well.

docs/CONTRIBUTING.md
IngmarBitter: ok.  Added an note that .editorconfig has limitations.

PR #67, opened 2020-05-20
HelloTexture ported from C++ DirectX-Graphics-Samples

IngmarBitter:
[Impl] HelloTexture from C++ samples ported to TerraFX. Compiles fine. Runtime needs fixing.


IngmarBitter:
recent commits where definitely an improvement and reflect what I learned on the other branches. 
HelloTriangle.cpp in the commit history is a convenient file to look at the relevant differences in the C++ reference. 
However, the extra debug setup is not giving me the D12 ERROR messages when something is going wrong and _device->CreateGraphicsPipelineState still crashes :(

samples/DirectX/D3D12/D3D12HelloTexture.cpp
```diff
@@ -0,0 +1,546 @@
+//*********************************************************
```
tannergooding, line 1: The C++ files should not be checked in, they provide no use in the actual codebase and should just be referenced instead (as the samples have already been doing).

samples/DirectX/D3D12/HelloTexture12.cs
```diff
@@ -0,0 +1,929 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
```
tannergooding, line 1: I can't comment on the actual file, but `HelloOutput.png` shouldn't be checked in. Checking in binary files generally isn't a good idea and while a few small sample pictures might be okay, a 60kb file isn't ideal and is a lot larger than would be liked.

I'd rather not check it in at all and look at alternatives for hosting any sample pictures.

IngmarBitter: Ah OK. I'll remove the reference to it. It's not essential.

samples/DirectX/D3D12/D3D12HelloTexture.cpp
IngmarBitter: removed it and the reference to it.

PR #82, opened 2020-06-02
Regenerating bindings using function pointers

tannergooding:
This regenerates all bindings using the new function pointers feature. Notably this currently uses `stdcall` as the calling convention but would ideally use `unmanaged` when it is supported.


tannergooding:
This is a draft PR as the official Roslyn nightly builds don't currently have the function-pointers feature yet and several of the tests don't currently pass due to:
> System.ArgumentException : Type 'TerraFX.Interop.IWICStream+Vtbl' cannot be marshaled as an unmanaged structure; no meaningful size or offset can be computed.

tannergooding:
I've temporarily dropped netcoreapp3.1 support until the dotnet/runtime issue can either be resolved or rejected.

PR #88, opened 2020-06-09
Add DXDIAG API from WinSDK

john-h-k:
(Also uppercases the DXCAPI GUIDs)


sources/Interop/Windows/um/dxdiag/_DXDIAG_INIT_PARAMS.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+// Ported from um/dxdiag.h in the Windows SDK for Windows 10.0.19041.0
+// Original source is Copyright © Microsoft. All rights reserved.
+
+namespace TerraFX.Interop
+{
+    public unsafe partial struct _DXDIAG_INIT_PARAMS
```
tannergooding, line 8: These should be adjusted to use the typedef name, `DXDIAG_INIT_PARAMS`

PR #101, opened 2020-07-14
Add console APIs

reflectronic:
See: 
https://docs.microsoft.com/en-us/windows/console/console-functions
https://docs.microsoft.com/en-us/windows/win32/api/wslapi/


generation/um/wincon/generate.rsp
```diff
@@ -0,0 +1,16 @@
+@../../settings.rsp
+@../../remap.rsp
+--file
+um-wincon.h
+--output
+../../../sources/Interop/Windows/um/wincon
+--test-output
+../../../tests/Interop/Windows/um/wincon
+--traverse
+C:/Program Files (x86)/Windows Kits/10/Include/10.0.19041.0/um/consoleapi.h
```
tannergooding, line 10: For TerraFX, we should only be traversing a single file from each `generate.rsp`

So separate `generate.rsp` are needed for `consoleapi.h`, etc

generation/um/wslapi/generate.rsp
```diff
@@ -0,0 +1,22 @@
+@../../settings.rsp
+@../../remap.rsp
+--exclude
+operator^
```
tannergooding, line 4: TerraFX is generally doing exact match exclusions to help clarify what is being dropped. If you add `--config log-exclusions` it will give you the full method name of anything that was partially matched.

sources/Interop/Windows/um/wincon/Windows.cs
```diff
@@ -0,0 +1,372 @@
...
+// Original source is Copyright © Microsoft. All rights reserved.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace TerraFX.Interop
+{
+    public static unsafe partial class Windows
+    {
+        [DllImport("kernel32", EntryPoint = "AllocConsole", ExactSpelling = true)]
```
tannergooding, line 13: There are `SetLastError` entries missing form these. @nathan-alden-sr actually wrote a program to help find these, but I can't seem to find it anymore.

sources/Interop/Windows/um/wslapi/Windows.cs
```diff
@@ -0,0 +1,41 @@
...
+// Original source is Copyright © Microsoft. All rights reserved.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace TerraFX.Interop
+{
+    public static unsafe partial class Windows
+    {
+        [DllImport("api-ms-win-wsl-api-l1-1-0", EntryPoint = "WslIsDistributionRegistered", ExactSpelling = true)]
```
tannergooding, line 13: This might be a docs issue, I think the actual assembly is meant to be `wslapi`

reflectronic: I believe that they are just recommending that you go through the API Set, but I can change it to `wslapi`

generation/um/wincon/generate.rsp
reflectronic: That is slightly unfortunate, since I was hoping that I could just go through `wincon.h` once. I'll split it into multiple directories

generation/um/wslapi/generate.rsp
```diff
@@ -0,0 +1,22 @@
+@../../settings.rsp
+@../../remap.rsp
+--exclude
+operator|(WSL_DISTRIBUTION_FLAGS, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS
+operator|=(WSL_DISTRIBUTION_FLAGS &, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS &
+operator&(WSL_DISTRIBUTION_FLAGS, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS
+operator&=(WSL_DISTRIBUTION_FLAGS &, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS &
+operator~(WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS
+operator^(WSL_DISTRIBUTION_FLAGS, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS
+operator^=(WSL_DISTRIBUTION_FLAGS &, WSL_DISTRIBUTION_FLAGS):WSL_DISTRIBUTION_FLAGS &
```
tannergooding, line 10: This should be fine to remove now, `exclude-enum-operators` gets it automatically
```suggestion
```

PR #102, opened 2020-07-15
Add windows.ui.composition.interop.h

reflectronic:
See: https://docs.microsoft.com/en-us/windows/win32/api/windows.ui.composition.interop/


generation/winrt/windows.ui.composition.interop/generate.rsp
```diff
@@ -0,0 +1,23 @@
+@../../settings.rsp
+@../../remap.rsp
+--exclude
+IDesktopWindowTarget
```
tannergooding, line 4: Why did this one need exclusion?

tannergooding: Looks like its an empty interface; which should be automatically handled now.

PR #106, opened 2020-07-15
Add windows.graphics.directx.direct3d11.interop.h

reflectronic:
See: https://docs.microsoft.com/en-us/windows/win32/api/windows.graphics.directx.direct3d11.interop/


sources/Interop/Windows/um/windows.graphics.directx.direct3d11.interop/Windows.Manual.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Runtime.InteropServices;
+
+namespace TerraFX.Interop
+{
+    public static unsafe partial class Windows
+    {
+        public static readonly Guid IID_IDirect3DDxgiInterfaceAccess = new Guid(0xA9B3D012, 0x3DF2, 0x4EE3, 0xB8, 0xD1, 0x86, 0x95, 0xF4, 0x57, 0xD3, 0xC1);
+    
+        public static IInspectable* CreateDirect3DDevice(IDXGISurface* dxgiSurface)
```
tannergooding, line 15: Is there an exclusion missing for this or did ClangSharp ignore it due to __cplusplus_winrt not being defined?

reflectronic: Yes, this is ignored since it's under that define. Clang would have choked on the C++/CX if it was being processed.

PR #136, opened 2020-09-14
Add schannel.h

reflectronic:
See: https://docs.microsoft.com/en-us/windows/win32/api/schannel/


generation/remap.rsp
```diff
@@ -1907,6 +1909,7 @@ _HID_DRIVER_CONFIG=HID_DRIVER_CONFIG
 _HID_XFER_PACKET=HID_XFER_PACKET
 _HISTOGRAM_BUCKET=HISTOGRAM_BUCKET
 _HMAC_Info=HMAC_Info
+_HMAPPER**=void**
```
tannergooding, line 1912: Was this needed rather than doing `_HMAPPER*=void*`?

sources/Interop/Windows/um/schannel/SCHANNEL_CERT_HASH_STORE.cs
```diff
@@ -0,0 +1,25 @@
...
+    public unsafe partial struct SCHANNEL_CERT_HASH_STORE
+    {
+        [NativeTypeName("DWORD")]
+        public uint dwLength;
+
+        [NativeTypeName("DWORD")]
+        public uint dwFlags;
+
+        [NativeTypeName("HCRYPTPROV")]
+        public nuint hProv;
```
tannergooding, line 17: Just noting for myself, `HCRYPTPROV` is a typedef to `ULONG_PTR`, it isn't defined like a normal Win32 handle.

tannergooding (approved):

PR #145, opened 2020-10-01
Merge TerraFX.Samples.WinForms back into the main solution

tannergooding:
This appears to work now that the main SDK can be used to build WinForms projects.


tannergooding:
Looks like building exes fail, but building libraries succeeds...

PR #151, opened 2020-11-15
ComPtr<T> improvements, new __uuidof<T> macro

Sergio0694:
This PR includes a few tweaks and improvements to the `ComPtr<T>` type, and a port of the `__uuidof<T>` macro.

🆕 Added a port of the `__uuidof()` macro, as a new static `__uuidof<T>()` API in the `Windows` class. This makes it easy to grab a `Guid*` value for a given type `T`, and similarly to the original macro it also can be used to assign a `Guid` value directly to a local variable. This works through a special `UuidOfType` type that wraps a `Guid*` pointer and adds implicit conversions when needed. This also indirectly speeds up a couple of `ComPtr<T>` APIs as the reflection done in `typeof(T).GUID` is no longer needed. Tier 1 JIT should also be able to completely replace the code with just a hardcoded address, but even without that it should be better than `typeof(T).GUID` every time. Also, this new API simplifies a lot having to grab a `rIID` value to use in various APIs.

🆕 Also added some utility `As` and `CopyTo` overloads.

Diff for the API surface:

```diff
 namespace TerraFX.Interop
 {
    public static class Windows
    {
+        public static unsafe UuidOfType __uuidof<T>() where T : unmanaged;

+        public readonly unsafe ref struct UuidOfType
+        {
+            public static implicit operator Guid(UuidOfType guid);
+            public static implicit operator Guid*(UuidOfType guid);
+        }
    }

     public unsafe struct ComPtr<T> : IDisposable
         where T : unmanaged
     {
         public ComPtr(T* other);
         public ComPtr(ComPtr<T> other);

         public static implicit operator ComPtr<T>(T* other);
         public static implicit operator T*(ComPtr<T> other);

         public readonly int As<U>(ComPtr<U>* p) where U : unmanaged;
+        public readonly int As<U>(ref ComPtr<U> other) where U : unmanaged;
         public readonly int AsIID(Guid* riid, ComPtr<IUnknown>* other);
+        public readonly int AsIID(Guid* riid, ref ComPtr<IUnknown> other);
         public void Attach(T* other);
         public T* Detach();
         public readonly int CopyTo(T** ptr);
+        public readonly int CopyTo(ComPtr<T>* p);
+        public readonly int CopyTo(ref ComPtr<T> other);
         public readonly int CopyTo<U>(U** ptr) where U : unmanaged;
+        public readonly int CopyTo<U>(ComPtr<U>* p) where U : unmanaged;
+        public readonly int CopyTo<U>(ref ComPtr<U> other) where U : unmanaged;
         public readonly int CopyTo(Guid* riid, void** ptr);
+        public readonly int CopyTo(Guid* riid, ComPtr<IUnknown>* p);
+        public readonly int CopyTo(Guid* riid, ref ComPtr<IUnknown> other);
         public void Dispose();
         public readonly T* Get();
         public readonly T** GetAddressOf();
         public readonly ref T* GetPinnableReference();
         public T** ReleaseAndGetAddressOf();
         public uint Reset();
+        public void Swap(ComPtr<T>* r);
         public void Swap(ref ComPtr<T> other);
     }
 }
```

🚀 Improved codegen to `GetAddressOf` and `ReleaseAndGetAddressOf`. The code was previously using a `fixed` block, which resulted in poor codegen as the JIT isn't currently able to optimize that away. Fixed that using `Unsafe` APIs:

```asm
; Old codegen
ComPtr`1[[System.Guid, System.Private.CoreLib]].GetAddressOf_Old()
    L0000: push rax
    L0001: xor eax, eax
    L0003: mov [rsp], rax
    L0007: mov [rsp], rcx
    L000b: mov rax, rcx
    L000e: add rsp, 8
    L0012: ret

; New codegen
ComPtr`1[[System.Guid, System.Private.CoreLib]].GetAddressOf_New()
    L0000: mov rax, rcx
    L0003: ret
```

📖 Minor code tweaks, added complete XML docs to the type


sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <summary>
+        /// Retrieves the GUID of of a specified type.
+        /// </summary>
+        /// <typeparam name="T">The type to retrieve the GUID for.</typeparam>
+        /// <returns>
+        /// A <see cref="GUID"/> value wrapping a pointer to the GUID data for the
+        /// input type. This value can be either converted to a <see cref="Guid"/>
+        /// pointer, or implicitly assigned to a <see cref="Guid"/> value.
+        /// </returns>
+        [Pure]
```
tannergooding, line 24: We don't use `Pure` anywhere else in the codebase.
```suggestion
```

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        /// Retrieves the GUID of of a specified type.
+        /// </summary>
+        /// <typeparam name="T">The type to retrieve the GUID for.</typeparam>
+        /// <returns>
+        /// A <see cref="GUID"/> value wrapping a pointer to the GUID data for the
+        /// input type. This value can be either converted to a <see cref="Guid"/>
+        /// pointer, or implicitly assigned to a <see cref="Guid"/> value.
+        /// </returns>
```
tannergooding, line 23: The rest of the repo, where applicable, uses single line summaries and doesn't do line wrapping.
```suggestion
        /// <summary>Retrieves the GUID of of a specified type.</summary>
        /// <typeparam name="T">The type to retrieve the GUID for.</typeparam>
        /// <returns>A <see cref="GUID"/> value wrapping a pointer to the GUID data for the input type. This value can be either converted to a <see cref="Guid"/> pointer, or implicitly assigned to a <see cref="Guid"/> value.</returns>
```

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return new(UUID<T>.RIID);
+        }
+
+        /// <summary>
+        /// A proxy type that wraps a pointer to GUID data. Values of this type can be implicitly
+        /// converted to and assigned to <see cref="Guid"/>* or <see cref="Guid"/> parameters.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public readonly unsafe ref struct GUID
```
tannergooding, line 37: I don't like calling this `GUID`. `GUID` is an existing native type and so I'd rather give this a more descriptive name that identifies its a helper. Maybe `UuidOfType` or similar.

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,89 @@
...
+            /// <param name="guid">The input <see cref="GUID"/> instance to read data for.</param>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static implicit operator Guid(GUID guid) => *guid.riid;
+
+            /// <summary>
+            /// Returns the <see cref="Guid"/>* pointer to the GUID buffer for a given <see cref="GUID"/> instance.
+            /// </summary>
+            /// <param name="guid">The input <see cref="GUID"/> instance to read data for.</param>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static implicit operator Guid*(GUID guid) => guid.riid;
```
tannergooding, line 59: What does the codegen look like for this and the other implicit operator?

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
+    /// to keep the reference counts properly tracked if you do store <see cref="ComPtr{T}"/> instances on the heap.
+    /// </remarks>
+    public unsafe struct ComPtr<T> : IDisposable
         where T : unmanaged
     {
-        private T* ptr_;
+        /// <summary>
+        /// The raw pointer to a COM object, if existing.
+        /// </summary>
+        private T* pointer;
```
tannergooding, line 28: This should remain `ptr_`. It matches the name as taken from the original source.

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
 
+        /// <summary>
+        /// Converts a raw pointer to a new <see cref="ComPtr{T}"/> instance and increments the ref count.
+        /// </summary>
+        /// <param name="other">The raw pointer to wrap.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static implicit operator ComPtr<T>(T* other)
         {
-            return new ComPtr<T>(other);
+            return new(other);
```
tannergooding, line 59: I don't like this syntax and likely won't be using it in TerraFX. It hurts readability.

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
Sergio0694: Right, removed in 294d7350829393cae965489877048b712cf409d7 🙂

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: Reverted in 650cef56939beacfd47c0eccfb75e500466f0f88, didn't know you wanted to keep the same name there 😄

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: Agreed it shouldn't always be used by default, I just figured it was fine in very small snippets like this.
Happy to revert this if you prefer though, done in b90b27e840a2b99fdaa552b9dd7f17385d0b056f 😊

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
Sergio0694: I'd say it looks quite nice 😄
Plus there's the fact that the tiered JIT could even elide the static constructor check entirely.
But other than this there's basically just the call to the static constructor (which we would've had anyway since `Windows` has statically initialized fields as well), and then all the rest is completely elided by the JIT.

<details>
  <summary><b>Guid* converter (click to expand):</b></summary>
  <br/>

```csharp
public static unsafe Guid* GetAddr()
{
    return __uuidof<ID3D12SomeType>();
}
```
```asm
sub      rsp, 40
mov      rcx, 0xD1FFAB1E
xor      edx, edx
call     CORINFO_HELP_CLASSINIT_SHARED_DYNAMICCLASS
mov      rax, 0xD1FFAB1E
mov      rax, qword ptr [rax]
add      rsp, 40
ret
```
</details>

<details>
  <summary><b>Guid converter (click to expand):</b></summary>
  <br/>

```csharp
public static unsafe Guid GetGuid()
{
    return Windows.__uuidof<ID3D12Test>();
}
```
```asm
push     rsi
sub      rsp, 32
vzeroupper 
mov      rsi, rcx
mov      rcx, 0xD1FFAB1E
xor      edx, edx
call     CORINFO_HELP_CLASSINIT_SHARED_DYNAMICCLASS
mov      rax, 0xD1FFAB1E
mov      rax, qword ptr [rax]
vmovdqu  xmm0, xmmword ptr [rax]
vmovdqu  xmmword ptr [rsi], xmm0
mov      rax, rsi
add      rsp, 32
pop      rsi
ret  
```
</details>

In particular for this as we have a direct comparison, the codegen is slightly smaller than just suing static GUIDs from the `Windows` class, but of course this approach is still much easier to use and less error prone.

<details>
  <summary><b>Accessing Windows.IID_Foo (click to expand):</b></summary>
  <br/>

```csharp
public static unsafe Guid GetGuid()
{
    return Windows.IID_Foo;
}
```
```asm
push     rsi
sub      rsp, 32
vzeroupper 
mov      rsi, rcx
mov      rcx, 0xD1FFAB1E
mov      edx, 5
call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
mov      rax, 0xD1FFAB1E
mov      rax, gword ptr [rax]
vmovdqu  xmm0, xmmword ptr [rax+8]
vmovdqu  xmmword ptr [rsi], xmm0
mov      rax, rsi
add      rsp, 32
pop      rsi
ret   
```
</details>

Considering that here we also have the prologue/epilogue that'd just go away once this method is just inlined 🚀

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
Sergio0694: Sounds good! Done in https://github.com/terrafx/terrafx.interop.windows/pull/151/commits/8702e3d0f93e5dbd69df9ff7c7acc64b63fd1f22 👍

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,81 @@
...
+        public static unsafe UuidOfType __uuidof<T>()
+            where T : unmanaged
+        {
+            return new UuidOfType(UUID<T>.RIID);
+        }
+
+        /// <summary>
+        /// A proxy type that wraps a pointer to GUID data. Values of this type can be implicitly
+        /// converted to and assigned to <see cref="Guid"/>* or <see cref="Guid"/> parameters.
+        /// </summary>
```
tannergooding, line 27: I had only called out the doc comment syntax mismatch on the type, but it applies to all added doc comments.

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
```diff
@@ -0,0 +1,81 @@
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public readonly unsafe ref struct UuidOfType
+        {
+            private readonly Guid* riid;
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
tannergooding, line 33: We shouldn't need `AggressiveInlining` on these. They are extremely small and the JIT should handle it already.

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
         private T* ptr_;
 
+        /// <summary>
+        /// Creates a new <see cref="ComPtr{T}"/> instance from a raw pointer and increments the ref count.
+        /// </summary>
+        /// <param name="other">The raw pointer to wrap.</param>
         public ComPtr(T* other)
         {
             ptr_ = other;
+
```
tannergooding, line 37: Unnecessary newline was inserted here and a few other places in the file.
```suggestion
```

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
+        /// Converts the current object reference to type <typeparamref name="U"/> and assigns that to a target <see cref="ComPtr{T}"/> value.
+        /// </summary>
+        /// <typeparam name="U">The interface type to use to try casting the current COM object.</typeparam>
+        /// <param name="other">A reference to the target <see cref="ComPtr{T}"/> value to write to.</param>
+        /// <returns>The result of <see cref="IUnknown.QueryInterface"/> for the target type <typeparamref name="U"/>.</returns>
+        /// <remarks>This method will automatically release the target COM object pointed to by <paramref name="other"/>, if any.</remarks>
+        public readonly int As<U>(ref ComPtr<U> other)
+            where U : unmanaged
+        {
+            fixed (ComPtr<U>* p = &other)
```
tannergooding, line 95: The  `pin` could be avoided here by using a local and then internally assigning `other.ptr_`.

tannergooding: Same with other cases like https://github.com/terrafx/terrafx.interop.windows/pull/151/files#diff-a1a828334f363e63b93afdd4a19f4f6ca865a85ceb9411137ecc4073be026014R122

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
+        /// <summary>
+        /// Increments the reference count for the current COM object, if any, and copies its address to a target <see cref="ComPtr{T}"/>.
+        /// </summary>
+        /// <param name="other">The target reference to copy the address of the current COM object to.</param>
+        /// <returns>This method always returns <see cref="S_OK"/>.</returns>
+        public readonly int CopyTo(ref ComPtr<T> other)
+        {
+            InternalAddRef();
+
+            fixed (ComPtr<T>* p = &other)
```
tannergooding, line 196: This pin could also be elided with the right helper. The address isn't actually being used, so releasing the value and assigning the field directly is probably better.

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
+                return ((IUnknown*)ptr_)->QueryInterface(riid, (void**)p->ReleaseAndGetAddressOf());
+            }
+        }
+
+        /// <inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Dispose()
         {
-            InternalRelease();
+            T* pointer = ptr_;
```
tannergooding, line 280: Why was this changed from using `InternalRelease()`?

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
```diff
@@ -4,150 +4,401 @@
...
-                ((IUnknown*)(ptr_))->AddRef();
+                ((IUnknown*)pointer)->AddRef();
             }
         }
 
+        // Decrements the reference count for the current COM object, if any
         private uint InternalRelease()
         {
-            uint @ref = 0;
+            uint referenceCount = 0;
```
tannergooding, line 390: The names here were matching the ones used in native code. It would be good to keep them where possible. Same for other places as well where changes have been made to the code.

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
Sergio0694: Ah, right. Fixed in b358ae9c1feade034dd730f43e777d6878e8ef18 🙂

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: But they looked so nice 😛
Fixed in b2a4be644512cb4da79f1df6c2c796a294b47b92!

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: I see, that makes perfect sense 😄
Reverted those changes in 3f93d6eea94a5528d3931543689c401c03289d07.

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: Removed all extra fixed blocks in e10f572606582cdb1e485423992a813aecd2bba8 🚀

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: Same as below, removed in e10f572606582cdb1e485423992a813aecd2bba8.

sources/Interop/Windows/winrt/wrl/client/ComPtr`1.Manual.cs
Sergio0694: Here we don't return the ref count so thought doing this manually and simplified could help the JIT with inlining?

sources/Interop/Windows/shared/uuids/Windows.Manual.cs
Sergio0694: Removed in 65b7f00150d78359e243ead4b46138d2c9624914.

tannergooding (approved):

PR #160, opened 2021-01-04
Add packaging and query APIs for app packages

reflectronic:
See:
https://docs.microsoft.com/en-us/windows/win32/api/appxpackaging/
https://docs.microsoft.com/en-us/windows/win32/api/appmodel/

There is a doc article [How to programmatically sign an app package](https://docs.microsoft.com/en-us/windows/win32/appxpkg/how-to-programmatically-sign-a-package) which describes some crypto APIs which do not come with the Windows SDK headers. I copied the definitions from the article into `other/mssign32/mssign32.h` to generate definitions for those APIs as well. Code samples and snippets in the Win32 docs are MIT licensed, and the header.txt links to the license file in the docs repository.


generation/other/mssign32/header.txt
```diff
@@ -0,0 +1,4 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+// Ported from https://docs.microsoft.com/en-us/windows/win32/appxpkg/how-to-programmatically-sign-a-package
+// Original source is Copyright © Microsoft. All rights reserved. License information available at https://github.com/MicrosoftDocs/win32/blob/docs/LICENSE-CODE
```
tannergooding, line 4: I'd prefer this explicitly state the license at the time it was ported, rather than relying on a potentially updatable url:
```suggestion
// Original source is Copyright © Microsoft. All rights reserved. Licensed under the MIT License (MIT).
```

tannergooding: Same for the other files, changes otherwise LGTM.

tannergooding (dismissed):

tannergooding (approved):

PR #161, opened 2021-01-05
Add XmlLite

saucecontrol:
This adds generation for `xmllite.h`

See: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms752872(v=vs.85)


generation/um/xmllite/generate.rsp
```diff
@@ -0,0 +1,22 @@
+@../../settings.rsp
+@../../remap.rsp
+--exclude
+IID_IXmlReader
+IID_IXmlWriter
+IID_IXmlResolver
+_DtdProcessing_Last
```
tannergooding, line 7: The `_Last` enum values are generally not excluded from bindings as they are still useful for various checks/asserts.

tannergooding (approved):

PR #163, opened 2021-01-11
Add various headers for core Windows Runtime functionality

reflectronic:
This PR adds various headers which relate to core Windows Runtime functionality. Other headers which fall into this category exist, but they depend on other larger headers which will be contributed in separate patches.

~~`roparameterizediid.h` in particular has some funky C++ helpers. The C# translation is slightly sketchy, and does some type punning to work around `UnmanagedCallersOnly` restrictions with generic types and methods: https://github.com/reflectronic/terrafx.interop.windows/blob/b3a33c08162b0eb40721e1c01690765763318a64/sources/Interop/Windows/winrt/roparameterizediid/Locator.Manual.cs It works properly in practice (at least, on x64), so I believe it is written correctly.~~ Update: We talked offline and decided to change the design.


generation/remap.rsp
```diff
@@ -3223,6 +3225,7 @@ __AnonymousBase_WinUser_L1173_C46=Base
 __AnonymousBase_WinUser_L13554_C43=Base
 __AnonymousBase_WinUser_L13558_C43=Base
 __AnonymousEnum_d2d1_L89_C1=D2D1_INTERPOLATION_MODE_DEFINITION
+__AnonymousRecord_roapi_L45_C9*=IntPtr
```
tannergooding, line 3228: Why does this need to be an `IntPtr`?

sources/Interop/Windows/winrt/roapi/Windows.Manual.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (SUCCEEDED(hr))
+            {
+                if (typeof(T).GUID == typeof(IInspectable).GUID)
+                {
+                    *instance = (T*) pInspectable;
+                }
+                else
+                {
+                    Guid iid = typeof(T).GUID;
+                    hr = pInspectable->QueryInterface(&iid, (void**) instance);
```
tannergooding, line 29: Since this is manual, you can use the `__uuidof` helper method

generation/remap.rsp
reflectronic: The definition for this struct in `roapi.h` is 
```cpp
typedef struct {} *RO_REGISTRATION_COOKIE;
```

sources/Interop/Windows/winrt/roparameterizediid/Locator.Manual.cs
```diff
@@ -0,0 +1,36 @@
...
+        void** lpVtbl;
+        delegate*<ushort*, IRoSimpleMetaDataBuilder*, int> _fn;
+
+        public _Locator(delegate*<ushort*, IRoSimpleMetaDataBuilder*, int> fn)
+        {
+            lpVtbl = computedVtbl;
+            _fn = fn;
+        }
+
+        static void** computedVtbl;
```
tannergooding, line 23: nit: These should all include `private`. Likewise, it would be "nice" to keep the `beforefieldinit` which just involves creating a helper method `private static void** CreateVtbl()`

tannergooding (approved):

sources/Interop/Windows/winrt/roparameterizediid/Locator.Manual.cs
reflectronic: Addressed in #165

PR #176, opened 2021-02-14
Add initial set of Windows Runtime APIs from ABI headers

reflectronic:
This adds the set of Windows Runtime APIs that are needed to launch a UWP app. The transitive closure of namespaces used by `Windows.UI.Core` and `Windows.ApplicationModel.Core` is pretty big, so there are a lot of seemingly unrelated namespaces in this PR. I tried to cull away some less relevant APIs by remapping things that would require pulling in another header to `IInspectable`.


rickbrew:
Related to #306 :)

PR #177, opened 2021-02-28
Add errhandlingapi.h

john-h-k:
Adds `um/errhandlingapi.h`

### DIFF
--- a/sources/Interop/Windows/um/errhandlingapi/Windows.cs
+++ b/sources/Interop/Windows/um/errhandlingapi/Windows.cs
@@ -1,0 +1,77 @@
+// Copyright © Tanner Gooding and Contributors. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+// Ported from um/errhandlingapi.h in the Windows SDK for Windows 10.0.19041.0
+// Original source is Copyright © Microsoft. All rights reserved.
+
+using System.Runtime.InteropServices;
+
+namespace TerraFX.Interop
+{
+    public static unsafe partial class Windows
+    {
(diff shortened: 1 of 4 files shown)


sources/Interop/Windows/um/errhandlingapi/Windows.cs
```diff
@@ -0,0 +1,77 @@
...
+        [DllImport("kernel32", ExactSpelling = true)]
+        [return: NativeTypeName("LPTOP_LEVEL_EXCEPTION_FILTER")]
+        public static extern delegate* unmanaged<EXCEPTION_POINTERS*, int> SetUnhandledExceptionFilter([NativeTypeName("LPTOP_LEVEL_EXCEPTION_FILTER")] delegate* unmanaged<EXCEPTION_POINTERS*, int> lpTopLevelExceptionFilter);
+
+        [DllImport("kernel32", ExactSpelling = true)]
+        [return: NativeTypeName("DWORD")]
+        public static extern uint GetLastError();
+
+        [DllImport("kernel32", ExactSpelling = true)]
+        public static extern void SetLastError([NativeTypeName("DWORD")] uint dwErrCode);
```
tannergooding, line 28: These APIs are problematic to expose and basically should never be called from .NET

The JIT can call many different native APIs in the background, even from say a GC interrupt, and cause `LastError` to become modified.

john-h-k: Noted, willfix

tannergooding:
This was resolved with a newer PR that handled some of the issues here, etc.

PR #225, opened 2021-10-30
Handling 64-bit and cross-architecture differences

tannergooding:


sources/Interop/Windows/um/wingdi/Windows.cs
```diff
@@ -1335,7 +1335,7 @@ public static unsafe partial class Windows
 
         [DllImport("opengl32", ExactSpelling = true, SetLastError = true)]
         [return: NativeTypeName("PROC")]
-        public static extern delegate* unmanaged<int> wglGetProcAddress([NativeTypeName("LPCSTR")] sbyte* param0);
+        public static extern IntPtr wglGetProcAddress([NativeTypeName("LPCSTR")] sbyte* param0);
```
rickbrew, line 1338: is this loss of the delegate type intentional?

tannergooding: Yes. `PROC` (and `FARPROC`) is meant to be an opaque function pointer. The actual type isn't known and you need to cast it to the correct signature.

PR #228, opened 2021-10-31
Simplify the user experience around um/SetupAPI and um/shellapi

tannergooding:


sources/Interop/Windows/um/shellapi/SHELLEXECUTEINFOW.Manual.cs
```diff
@@ -0,0 +1,283 @@
...
+// Original source is Copyright © Microsoft. All rights reserved.
+
+using System.Runtime.InteropServices;
+
+namespace TerraFX.Interop
+{
+    [StructLayout(LayoutKind.Explicit)]
+    public unsafe partial struct SHELLEXECUTEINFOW
+    {
+        public static uint Size
```
rickbrew, line 13: maybe call this `SizeOfValue` ? or `ValueSize` ?

tannergooding: What about just `SizeOf`?

Issue #234, opened 2021-11-02
BufferedPaintMakeOpaque should return HRESULT


rickbrew:
It's a macro around BufferedPaintSetAlpha, which returns HRESULT. Easy to miss. https://docs.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-bufferedpaintsetalpha

> The BufferedPaintMakeOpaque macro, which is defined in uxtheme.h, sets alpha values to 255. It is typically used to call GDI to draw into a memory buffer and then to make it opaque in order to draw it on glass.

rickbrew:
Same for `SetWindowThemeNonClientAttributes`.

I'm not using either of these, btw.

Issue #201, opened 2021-10-12
Missing GET_X_LPARAM and GET_Y_LPARAM


rickbrew:
I went to go use these "macros" in my code and they're missing from TerraFX

`GET_X_LPARAM()` maps to `LOWORD()`, while `GET_Y_LPARAM()` maps to `HIWORD()`

rickbrew:
thanks!

PR #239, opened 2021-11-02
Ensuring transparent struct wrappers are used in manual defines and port shared/windowsx.h

tannergooding:
This resolves https://github.com/terrafx/terrafx.interop.windows/issues/234
This resolves https://github.com/terrafx/terrafx.interop.windows/issues/201

### DIFF
--- a/sources/Interop/Windows/Windows.cs
+++ b/sources/Interop/Windows/Windows.cs
@@ -12,10 +12,7 @@ namespace TerraFX.Interop
         /// <summary>Raised whenever a native library is loaded by TerraFX.Interop.Windows. Handlers can be added to this event to customize how libraries are loaded, and they will be used first whenever a new native library is being resolved.</summary>
         public static event DllImportResolver? ResolveLibrary;
 
-        /// <summary>Initializes the DLL resolver for the current assembly.</summary>
-        /// <returns>Just <see langword="null"/>, this method is only used to skip the static constructor and keep this type as <see langword="beforefieldinit"/>.</returns>
-        [ModuleInitializer]
-        internal static void SetDllImportResolver()
+        static Windows()
         {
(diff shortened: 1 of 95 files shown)


sources/Interop/Windows/Windows.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace TerraFX.Interop
+{
+    public static unsafe partial class Windows
+    {
+        /// <summary>Raised whenever a native library is loaded by TerraFX.Interop.Windows. Handlers can be added to this event to customize how libraries are loaded, and they will be used first whenever a new native library is being resolved.</summary>
+        public static event DllImportResolver? ResolveLibrary;
+
+        static Windows()
+        {
+            NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), OnDllImport);
```
rickbrew, line 17: You don't have to call this until there's an event subscription on `ResolveLibrary`, e.g.

`private static DllImportResolver? resolveLibrary;
public static DllImportResolver? ResolverLibrary
{
    add
    {
        ... do it here ...
    }
}
```

This helps ensure this code path isn't executed unless needed. And, this would certainly cause problems if anyone decides to do IL merging and also needs to call SetDllImportResolver() on the assembly that TerraFX is being merged into.

tannergooding: This has to be done up front because `SetDllImportResolver` needs to be done before the first P/Invoke call happens because a given library is resolved once and you can only have a single hook for a given assembly. This means if I don't set it up and first thing, it causes downstream issues if users try to set up their own or if I ever need to add default resolution.

Default resolution logic is required for some assemblies, like vulkan, because the name differs per OS, its also theoretically needed for cases like x3daudio or d3dcompiler to help avoid versioning issues because the assembly name can change across versions.

Issue #240, opened 2021-11-06
Expose DbgHelp.h


DaZombieKiller:
https://docs.microsoft.com/en-us/windows/win32/debug/debug-help-library
Comprised of `DbgHelp.h` and `minidumpapiset.h`. Also requires exposing `verrsrc.h` and `timezoneapi.h`.

AraHaan:
@DaZombieKiller you plan to do a follow-up pr for ``minidumpapiset.h`` soon?

DaZombieKiller:
@AraHaan At some point later I may get around to it, but I'm busy with other work currently so it won't be any time soon.

PR #241, opened 2021-11-06
[IMPL] Expose DbgHelp.h

DaZombieKiller:
Closes https://github.com/terrafx/terrafx.interop.windows/issues/240
This also exposes `verrsrc.h` and `timezoneapi.h` as they weren't previously exposed and are required by the DbgHelp API.
`minidumpapiset.h` currently has a number of issues, so I'll look at exposing it in a separate PR at a later time. I've still kept the type remappings for it in anticipation of that.


generation/um/minidumpapiset/generate.rsp
```diff
@@ -0,0 +1,19 @@
...
+--file
+um-minidumpapiset.h
+--output
+../../../sources/Interop/Windows/um/minidumpapiset
+--test-output
+../../../tests/Interop/Windows/um/minidumpapiset
+--traverse
+C:/Program Files (x86)/Windows Kits/10/Include/10.0.20348.0/um/minidumpapiset.h
+--exclude
+MiniDumpWriteDump
```
AraHaan, line 12: Uh I would not exclude MiniDumpWriteDump, instead I could suggest to route ``_MINIDUMP_CALLBACK_INFORMATION``, ``_MINIDUMP_CALLBACK_INPUT``, ``_MINIDUMP_THREAD_CALLBACK``, ``_MINIDUMP_THREAD_EX_CALLBACK`` which I think it takes in as a pointer to ``IntPtr`` or ``nint``?

tannergooding (approved):

PR #270, opened 2021-12-01
Add the primary ACL APIs (AclAPI.h)

JeremyKuhne:
Key thing here is being able to get effective rights. Includes `acccltrl.h` which has the necessary supporting defines.

Note that there really isn't a great way to group `ACTRL_*` defines any better. It gets particularly awkward as there are multiple `ACTRL_*` structs that overlap. :)

@tannergooding do you want me to be opening matching issues with these PRs?


tannergooding:
> do you want me to be opening matching issues with these PRs?

For smallish PRs it should be fine to just open a PR :)

tannergooding (approved):

PR #273, opened 2021-12-02
Add ntsecapi.h & lsalookup.h

### LINKED ISSUE #266: Add LSA related APIs in ntsecapi.h
These are necessary to check SID rights, logon sessions, etc.

https://docs.microsoft.com/en-us/windows/win32/api/ntsecapi/

Another useful API here is `LsaNtStatusToWinError` which allows converting `NTSTATUS` codes to Windows error codes.

JeremyKuhne:
Adds a number of key security APIs and types.

Fixes #266

I'm not quite sure what to do with some of the existing defines moving as `POLICY_` and `LOGON_` cross the automatic thresh-hold.

Also: There were a few defines for "function-like" macros that I've excluded, and a couple of undocumented APIs that aren't exposed in any dll exports that I can find.


JeremyKuhne:
Rebased

tannergooding:
> I'm not quite sure what to do with some of the existing defines moving as POLICY_ and LOGON_ cross the automatic thresh-hold.

It's generally fine/expected for the constants to move when the threshold is crossed. It will be source breaking only and there isn't much of a better option available given how large the Windows SDK is.

> Also: There were a few defines for "function-like" macros that I've excluded, and a couple of undocumented APIs that aren't exposed in any dll exports that I can find.

Sounds good. Function-like macros can't currently be automatically handled and undocumented APIs shouldn't be exposed.

tannergooding (approved):

PR #287, opened 2021-12-16
Change remapping for `UI_ANIMATION_KEYFRAME`

### LINKED ISSUE #286: `UI_ANIMATION_KEYFRAME` is defined incorrectly
The headers define it like this,

```c
typedef struct __MIDL___MIDL_itf_UIAnimation_0000_0002_0003
    {
    int _;
    }     *UI_ANIMATION_KEYFRAME;
```

TerraFX defines it as

```cs
struct UI_ANIMATION_KEYFRAME
{
    public int _;
}
```

It should really be `public int* _;` or something similar to a `HANDLE` wrapper.

The COM interface structs do use `UI_ANIMATION_KEYFRAME*` instead of `UI_ANIMATION_KEYFRAME`, so it's technically ABI-correct. However, it's quite confusing and difficult to work with.

rickbrew:
Fixes #286

~~Still doesn't seem to work quite right, the COM interface structs aren't being regenerated and still have the wrong indirection level~~


rickbrew:
Ready for review 👍🏼

tannergooding (approved):

tannergooding (approved):

Issue #288, opened 2021-12-17
Some missing HRESULT errors from winerror.h


rickbrew:
Ran across the `E_NOT_SUFFICIENT_BUFFER` while porting over my `IDWriteTextLayout` wrapper, and noticed the others were also missing

```c
#define E_NOT_SET                HRESULT_FROM_WIN32(ERROR_NOT_FOUND)
#define E_NOT_VALID_STATE        HRESULT_FROM_WIN32(ERROR_INVALID_STATE)
#define E_NOT_SUFFICIENT_BUFFER  HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER)
#define E_TIME_SENSITIVE_THREAD  HRESULT_FROM_WIN32(ERROR_TIME_SENSITIVE_THREAD)
#define E_NO_TASK_QUEUE          HRESULT_FROM_WIN32(ERROR_NO_TASK_QUEUE)
```

rickbrew:
Is this okay to close?

Issue #42, opened 2020-04-21
Consider exposing some sort of ThrowIfFailed(HRESULT) method


john-h-k:
### Proposals

One such is already exposed in HelperUtilities, although this is an internal utility type. Would be very useful for code which uses lower-level portions of the library where checking HRESULTs is a regular occurence

nathan-alden-sr:
I wrote an `HResult` type that does just that. See here: https://github.com/bouncy-box/bouncybox-vorpalengine-old/blob/master/src/Interop/HResult.cs

nathan-alden-sr:
I'm okay with Tanner consuming that type if he wants. One less thing for me to maintain. 😛

rickbrew:
Can this be closed?

PR #296, opened 2021-12-23
Ensure E_NOT_SET + friends are being generated and add a ThrowIfFailed extension to HRESULT

tannergooding:
This resolves https://github.com/terrafx/terrafx.interop.windows/issues/288 and resolves https://github.com/terrafx/terrafx.interop.windows/issues/42


sources/Interop/Windows/Windows/other/helper-types/Windows.Manual.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    [DoesNotReturn]
+    public static void ThrowExternalException(string methodName, int errorCode)
+    {
+        var message = string.Format("'{0}' failed with an error code of '{1}'", methodName, errorCode);
+        throw new ExternalException(message, errorCode);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void ThrowIfFailed(this HRESULT value, [CallerArgumentExpression("value")] string? valueExpression = null)
```
rickbrew, line 19: For my uses, this is problematic -- I have my own `ThrowOnError()` extension method that goes through an `ExceptionFactory`. The HRESULT is converted into a specific exception, whether that's `IOException`, `ArgumentException`, or library-specific exceptions like `PaintDotNet.Direct2D1.RecreateTargetException` or `Animation.IllegalReentrancyException`. 

It's problematic because my muscle memory is now such that I type `hr.T` and press tab, which currently auto-completes to `hr.ThrowOnError()`. However, now it might auto-complete to `ThrowIfFailed()`, and it will be difficult to catch these mistakes. I do have `catch`'s where I go for the specific exception, not just `ExternalException`. So this is very important for my code base, whether writing new code or maintaining old code.

If there was a way to install an exception factory into TerraFX, then I could plug mine in and remove my own exception factory. There are probably other ways to solve this, too.

An example:
```cs
public ITransformedGeometry CreateTransformedGeometry(IGeometry source, Matrix3x2Float transform)
{
    using ComPtr<ID2D1Geometry> spSource = default;
    HRESULT hr = ComObject.GetIUnknown(source, &spSource);
    hr.ThrowOnError();

    using ComPtr<ID2D1TransformedGeometry> spTransformedGeometry = default;
    hr = this.pD2D1Factory->CreateTransformedGeometry(
        spSource,
        (D2D_MATRIX_3X2_F*)&transform,
        spTransformedGeometry.GetAddressOf());
    hr.ThrowOnError();

    return ComObjectRef.CreateRef<ITransformedGeometry>(spTransformedGeometry)!;
}
```

tannergooding: I've dropped it being an extension method in favor of just being a static method here: https://github.com/terrafx/terrafx.interop.windows/pull/297

The alternative would have been for your `ThrowOnError` to be named the same (`ThrowIfFailed`) in which case it would be preferenced due to being a "closer" extension method

PR #298, opened 2021-12-26
Generating bindings for most of GDI+

tannergooding:
This replaces https://github.com/terrafx/terrafx.interop.windows/pull/208 and generates bindings for most of GDI+, the parts that are missing are namely the C++ surface area which needs more tweaking to get right.


AraHaan:
Will these get ported to branch main as well?

PR #341, opened 2023-02-26
Update TerraFX.Interop.Windows to support trimming the `SetDllImportResolver` call

tannergooding:
Notably this does not actually remove the `static constructor` as the trimmer currently leaves it as:
```csharp
static Windows()
{
  if (true)
    ;
}
```

Likewise, the field `private static readonly bool s_disableResolveLibraryHook` field cannot be consumed directly, doing so leaves the field, its initializer, and other "dead code". Accessing it through the property ends up working, however.

Attempting to remove the static constructor directly likewise does not work, we are left with:
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
static Windows()
{
  // ISSUE: unable to decompile the method.
}
```

### DIFF
--- a/samples/DirectX/TerraFX.Samples.DirectX.csproj
+++ b/samples/DirectX/TerraFX.Samples.DirectX.csproj
@@ -5,7 +5,6 @@
   <PropertyGroup>
     <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
   </PropertyGroup>
-  
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


samples/DirectX/TerraFX.Samples.DirectX.csproj
```diff
@@ -5,9 +5,13 @@
   <PropertyGroup>
     <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
   </PropertyGroup>
-
+
```
tannergooding, line 8:
```suggestion

```

PR #342, opened 2023-02-26
Make ResolveLibrary pay-for-play

jkotas:
As added benefit, this also removes static constructor trigger on the Windows static class and makes them faster (in AOT scenarios or when tiering is disabled)

Contributes to: https://github.com/dotnet/runtime/issues/82607

### DIFF
--- a/sources/Interop/Windows/Windows.cs
+++ b/sources/Interop/Windows/Windows.cs
@@ -9,51 +9,61 @@ namespace TerraFX.Interop.Windows;
 public static unsafe partial class Windows
 {
     /// <summary>Raised whenever a native library is loaded by TerraFX.Interop.Windows. Handlers can be added to this event to customize how libraries are loaded, and they will be used first whenever a new native library is being resolved.</summary>
-    public static event DllImportResolver? ResolveLibrary;
-
-    static Windows()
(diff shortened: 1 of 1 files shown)


tannergooding (changes requested):
I appreciate the PR, but this isn't one that I'll be taking.

As explained on dotnet/runtime#82607, the size savings (without also opting into multiple other runtime configuration knobs) is about 1kb trimmed and 4kb aot.

This code is quite a bit more verbose and will lose its benefit as soon as I end up generating bindings for any native library that does need specialized resolution, and there are a few of them particularly given the move of some components (like DirectX) to be more "agile" and provide out of band versions that can be pulled from places like NuGet.

The alternative I've put up in https://github.com/terrafx/terrafx.interop.windows/pull/341 works more generally. It causes no changes to the default experience for most users and works across all the `TerraFX.Interop.*` libraries. For the power users which are fine-tuning things, its a simple opt-in they can find in the documentation that will let them trim a little bit extra if this is the last thing rooting any reflection logic. 

Additionally, if you are a power user opting into the various runtime config switches to "fine tune" the size of your output, then opting into an additional TerraFX config switch is not a far jump and shouldn't present any additional burden to such users.

PR #343, opened 2023-02-27
Don't use NativeLibrary.Load unnecessarily

tannergooding:


sources/Interop/Windows/Windows.cs
```diff
@@ -15,7 +15,7 @@ static Windows()
     {
         if (!Configuration.DisableResolveLibraryHook)
         {
-            NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), OnDllImport);
+            NativeLibrary.SetDllImportResolver(typeof(Windows).Assembly, OnDllImport);
```
MichalStrehovsky, line 18: This will break things with reflection disabled mode, so that's a nice bonus from my perspective because I don't want people to use that :).

GetExecutingAssembly is an intrinsic that happens to work with reflection disabled.

tannergooding:
```suggestion
            NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), OnDllImport);
```

PR #356, opened 2023-06-29
Allow disabling unused analyzers

tannergooding:
This disables `Microsoft.Interop.JavaScript.JSImportGenerator` as it is unused but adding a considerable amount of build time (33% locally).


tannergooding:
This is mostly fixed in the latest versions of the SDK.

PR #386, opened 2024-01-21
Update Win32Application.cs

DeafMan1983:
Shouldn't default: because I have tested since I clicked Titlebar than maximized and unmaximized without default: cause It crashed since you wrote default: { ... }

It fixed without default: It works as well under Windows 10 ( latest build version )

Proof: I have tested with debug and release without default: { ... } = Works fine 100 %.

Thank you for accepting my solution!

### DIFF
--- a/samples/DirectX/Shared/Win32Application.cs
+++ b/samples/DirectX/Shared/Win32Application.cs
@@ -176,11 +176,8 @@ public static unsafe class Win32Application
                 pSample?.OnKeyUp((byte)wParam);
                 return 0;
             }
-
-            default:
-            {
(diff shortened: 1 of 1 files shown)


tannergooding:
There's no difference between the two pieces of code. `default: { }` is simply the case that is executed if no other entry is matched and if no `default` exists, then it executes the next statement following the `switch` block.

It only comes down to stylistic preference, and personally I prefer having it as part of the `switch` to make it clear that every case returns and there is no additional code that will execute.

DeafMan1983:
Ok thanks for understanding!

Issue #387, opened 2024-03-02
Add implicit conversion from RECT -> System.Drawing.Rectangle (System.Drawing.Primitives v8.0.0 from base shared framework).


AraHaan:
#### Description (optional)

I think an helpful implicit conversion to ``Rectangle`` would help with everything with APIs that uses ``RECT`` or returns ``RECT``.

#### Rationale

Why to System.Drawing.Rectangle? Because the type exists in System.Drawing.Primitives which is part of the base shared framework which means it is not locked down to ``Microsoft.WindowDesktop.App`` as it exists in the ``Microsoft.NETCore.App`` shared framework which means it is available everywhere for every operating system. With it existing as a reference to every project that references the framework without it disabled that means it is safe to use in TerraFX.Interop.Windows as it is a primitive type. Also having operators between the 2 will help with ``RECT`` <-> ``Rectangle`` conversions without things going wrong when the developer messes up conversions as the operators will take all the guess work out of it all.

#### Proposed API

```diff
namespace TerraFX.Interop.Windows;

public partial struct RECT
{
+    public static implicit operator Rectangle(RECT rectangle);
+    public static explicit operator RECT(Rectangle rectangle);
}
```

#### Drawbacks

I think an implicit conversion like this would have minimal drawbacks with an explicit cast from ``Rectangle`` -> ``RECT`` as well so people can also pass those to APIs without doing it the long way as well (it becomes problematic to remember what should go into the ``Left``, ``Right``, ``Top``, ``Bottom`` so operators that handles it for us could also help prevent bugs in people's code as well as convenience as well.

#### Alternatives

I have not though of any alternatives yet.

#### Other thoughts

Implementation would look something like this of the new operators:

```cs
namespace TerraFX.Interop.Windows;

public partial struct RECT
{
    public static implicit operator Rectangle(RECT rectangle)
        => Rectangle.FromLTRB(rectangle.left, rectangle.top, rectangle.right, rectangle.bottom);

    public static explicit operator RECT(Rectangle rectangle)
        => new // or new RECT()
        {
            // Obtained from inspecting 'Rectangle.FromLTRB(int, int, int, int)'.
            left = rectangle.X,
            top = rectangle.Y,
            right = rectangle.Width + rectangle.X,
            bottom = rectangle.Height + rectangle.Y,
        };
}
```

#### Discussions (optional)

[...]

PR #388, opened 2024-03-02
[IMPL] Add implicit conversion from RECT -> System.Drawing.Rectangle.

AraHaan:
## Proposal Implementation

Adds an implicit conversion operator from ``RECT`` -> ``Rectangle`` with an explicit operator from ``Rectangle`` back to ``RECT`` when the rectangle needs converted to an ``RECT`` for calls into Windows APIs that requires it. Also with operators handling these conversions programmers would worry less on if they got their conversion between ``RECT`` <-> ``Rectangle`` correct or not by eliminating all the guess work from it. Unit tests of the operators will come soon. As such I opened this pull request as a draft.

Implements/Fixes https://github.com/terrafx/terrafx.interop.windows/issues/387.

### DIFF
--- a/sources/Interop/Windows/Windows/shared/windef/RECT.cs
+++ b/sources/Interop/Windows/Windows/shared/windef/RECT.cs
@@ -5,6 +5,8 @@
 
 namespace TerraFX.Interop.Windows;
 
+using System.Drawing;
+
 /// <include file='RECT.xml' path='doc/member[@name="RECT"]/*' />
 public partial struct RECT
 {
@@ -23,4 +25,19 @@ public partial struct RECT
     /// <include file='RECT.xml' path='doc/member[@name="RECT.bottom"]/*' />
     [NativeTypeName("LONG")]
     public int bottom;
+
+    // type 'Rectangle' is from System.Drawing.Primitives which is from the base shared framework.
+    // as such these operators *should* be safe to include here.
+    public static implicit operator Rectangle(RECT rectangle)
+        => Rectangle.FromLTRB(rectangle.left, rectangle.top, rectangle.right, rectangle.bottom);
+
+    public static explicit operator RECT(Rectangle rectangle)
+        => new
+        {
+            // Obtained from inspecting 'Rectangle.FromLTRB(int, int, int, int)'.
+            left = rectangle.X,
+            top = rectangle.Y,
+            right = rectangle.Width + rectangle.X,
+            bottom = rectangle.Height + rectangle.Y,
+        };
 }
--- a/sources/Interop/Windows/Windows/shared/windef/RECT.cs
+++ b/sources/Interop/Windows/Windows/shared/windef/RECT.cs
@@ -28,11 +28,11 @@ public partial struct RECT
 
     // type 'Rectangle' is from System.Drawing.Primitives which is from the base shared framework.
     // as such these operators *should* be safe to include here.
-    public static implicit operator Rectangle(RECT rectangle)
+    public static explicit operator Rectangle(RECT rectangle)
         => Rectangle.FromLTRB(rectangle.left, rectangle.top, rectangle.right, rectangle.bottom);
 
     public static explicit operator RECT(Rectangle rectangle)
-        => new
+        => new RECT()
         {
             // Obtained from inspecting 'Rectangle.FromLTRB(int, int, int, int)'.
             left = rectangle.X,


sources/Interop/Windows/Windows/shared/windef/RECT.cs
```diff
@@ -23,4 +25,19 @@ public partial struct RECT
     /// <include file='RECT.xml' path='doc/member[@name="RECT.bottom"]/*' />
     [NativeTypeName("LONG")]
     public int bottom;
+
+    // type 'Rectangle' is from System.Drawing.Primitives which is from the base shared framework.
+    // as such these operators *should* be safe to include here.
+    public static implicit operator Rectangle(RECT rectangle)
+        => Rectangle.FromLTRB(rectangle.left, rectangle.top, rectangle.right, rectangle.bottom);
```
tannergooding, line 32: This isn't safe to be implicit, as it potentially overflows and thus is lossy.

sources/Interop/Windows/Windows/shared/windef/RECT.cs
```diff
@@ -23,4 +25,19 @@ public partial struct RECT
...
+
+    public static explicit operator RECT(Rectangle rectangle)
+        => new
+        {
```
tannergooding, line 36: nit: `new()` is one of my least favorite language features and I'm pretty sure it's blocked in the editorconfig, so I expect this will cause CI to fail.

tannergooding:
Same general feedback as given on the issue. I'm not particularly a fan of this, as its potentially lossy and brings in a new dependency.

This really seems like something where a user could provide extension methods for their own library to solve the need.

sources/Interop/Windows/Windows/shared/windef/RECT.cs
```diff
@@ -23,4 +25,19 @@ public partial struct RECT
...
     [NativeTypeName("LONG")]
     public int bottom;
+
+    // type 'Rectangle' is from System.Drawing.Primitives which is from the base shared framework.
+    // as such these operators *should* be safe to include here.
+    public static implicit operator Rectangle(RECT rectangle)
+        => Rectangle.FromLTRB(rectangle.left, rectangle.top, rectangle.right, rectangle.bottom);
+
+    public static explicit operator RECT(Rectangle rectangle)
+        => new
```
AraHaan, line 35:
```suggestion
        => new RECT()
```

sources/Interop/Windows/Windows/shared/windef/RECT.cs
```diff
@@ -23,4 +25,19 @@ public partial struct RECT
     /// <include file='RECT.xml' path='doc/member[@name="RECT.bottom"]/*' />
     [NativeTypeName("LONG")]
     public int bottom;
+
+    // type 'Rectangle' is from System.Drawing.Primitives which is from the base shared framework.
+    // as such these operators *should* be safe to include here.
+    public static implicit operator Rectangle(RECT rectangle)
```
AraHaan, line 31:
```suggestion
    public static explicit operator Rectangle(RECT rectangle)
```

rickbrew:
> Same general feedback as given on the issue. I'm not particularly a fan of this, as its potentially lossy and brings in a new dependency.
> 
> This really seems like something where a user could provide extension methods for their own library to solve the need.

I'm actually strongly opposed to this pull request. I do not think it belongs in TerraFX, and it would actually be a bug hazard in my own codebase. `RECT` and `Rectangle` are not interchangeable, they simply cannot represent the same values.

For instance, `Rectangle` cannot encode a rectangle with a width greater than `int.MaxValue`, while `RECT` can. Any `RECT` where `right - left` is greater than `int.MaxValue` will have silent overflows or worse. I have seen this sort of thing cause major problems in my own code base, it is not a theoretical problem. I have a lot of interop and conversions between "XYWH" rectangles (`Rectangle`, `WICRect`) and "LTRB" rectangles (`RECT`, `D2D_RECT`, etc). It is not something you can correctly handle with some implicit casting operators.

AraHaan:
> > Same general feedback as given on the issue. I'm not particularly a fan of this, as its potentially lossy and brings in a new dependency.
> > This really seems like something where a user could provide extension methods for their own library to solve the need.
> 
> I'm actually strongly opposed to this pull request. I do not think it belongs in TerraFX, and it would actually be a bug hazard in my own codebase. `RECT` and `Rectangle` are not interchangeable, they simply cannot represent the same values.
> 
> For instance, `Rectangle` cannot encode a rectangle with a width greater than `int.MaxValue`, while `RECT` can. Any `RECT` where `right - left` is greater than `int.MaxValue` will have silent overflows or worse. I have seen this sort of thing cause major problems in my own code base, it is not a theoretical problem. I have a lot of interop and conversions between "XYWH" rectangles (`Rectangle`, `WICRect`) and "LTRB" rectangles (`RECT`, `D2D_RECT`, etc). It is not something you can correctly handle with some implicit casting operators.

I guess I should have the operators check for if > ``int.MaxValue`` on ``right - left`` and if so, do something about it, Issue is what should be done in the case of ``right - left``? Throw, or do some other type of logic to make it work in a way that won't make Rectangle throw?

rickbrew:
Yup. If `right` is `int.MaxValue` and `left` is `int.MinValue`, then `width` should be `uint.MaxValue`, but that will not be representable as a (signed) `int` and will overflow. There are actually a substantial number of corner cases to handle when working with arbitrary rectangles. I still find bugs in my handling of `RECT` vs. `Rectangle` handling in Paint.NET, even after 20 years. They can be very subtle and difficult do diagnose.

I think @tannergooding has the right idea here. You should have your own utility or extension methods to handle the conversion, and it should not be baked into TerraFX.

The answer to "what should be done if ..." is going to be application specific. In some cases you want to throw an exception or do a no-op. In others you want to use a special value for "empty" or "logically infinite" rectangles. I think having operators baked into TerraFX would lock in one of these choices and make things a lot harder for anyone who needs the other ways of handling this. 

Put another way: there are too many ways to get this very wrong, and I don't see any way to get this right for all applications.

tannergooding:
Going to close this. Rick was able to reiterate the points I intended.

PR #403, opened 2024-12-19
Add Presentation.h and PresentationTypes.h (Win11 Composition Swapchain API)

rickbrew:
This looks right except that `Windows.CreatePresentationFactory` has a blank DLL name: `[DllImport("", ...)]`. It should be `dcomp.dll` as per https://learn.microsoft.com/en-us/windows/win32/api/presentation/nf-presentation-createpresentationfactory

_Edit:_ Fixed the DLL name for the `[DllImport]` on `CreatePresentationFactory`


rickbrew:
The unresolved `TaskDialogIndirect` is because there's neither a manifest nor an activation context that specifies to use version 6 of the DLL instead of 5.82. Normal stuff.

generation/Windows/um/Presentation/generate.rsp
```diff
@@ -0,0 +1,21 @@
...
+../../../../tests/Interop/Windows/Windows/um/Presentation
+--traverse
+C:/Program Files (x86)/Windows Kits/10/Include/10.0.26100.0/um/Presentation.h
+--with-librarypath
+CreatePresentationFactory=dcomp
+--with-using
+CompositionFrameDisplayInstance=TerraFX.Interop.DirectX
+ICompositionFramePresentStatistics=TerraFX.Interop.DirectX
+IPresentationSurface=TerraFX.Interop.DirectX
+IPresentationSurface2=TerraFX.Interop.DirectX
```
tannergooding, line 21: You should be able to get these automatic by adding the type it uses to https://github.com/terrafx/terrafx.interop.windows/blob/main/generation/remap-namespaces.rsp, which should just tell the system that `DXGI_COLOR_SPACE_TYPE` is in `TerraFX.Interop.DirectX`

tannergooding:
> ResolveDllImportTests is probably failing because the CI server isn't running on a build of Windows that has these APIs? A whole bunch of other APIs are failing the same test, too.

You should be able to add the entries to the handler in https://github.com/terrafx/terrafx.interop.windows/blob/main/tests/Interop/Windows/ResolveDllImportTests.cs#L53

There's an existing grouping for cases that simply aren't on server, but if its instead relegated to a particular version of Windows then you can place it in a separate grouping with the relevant guard.

tannergooding:
> The unresolved TaskDialogIndirect is because there's neither a manifest nor an activation context that specifies to use version 6 of the DLL instead of 5.82. Normal stuff.

That is already be getting handled via https://github.com/terrafx/terrafx.interop.windows/blob/main/tests/Interop/Windows/ResolveDllImportTests.cs#L218 and warning, not erroring.

The only actual failure is `CreatePresentationFactory`

rickbrew:
> > The unresolved TaskDialogIndirect is because there's neither a manifest nor an activation context that specifies to use version 6 of the DLL instead of 5.82. Normal stuff.
> 
> That is already be getting handled via https://github.com/terrafx/terrafx.interop.windows/blob/main/tests/Interop/Windows/ResolveDllImportTests.cs#L218 and warning, not erroring.
> 
It should be possible to test the `comctl32.dll` exports without a manifest by using an activation context. Might not be worth the trouble though.

rickbrew:
`CreatePresentationFactory` does indeed work :)

image

```cs
using static TerraFX.Interop.Windows.Windows;

unsafe void Main()
{
    using ComPtr<IDXGIFactory7> spDxgiFactory = default;
    HRESULT hr = DirectX.CreateDXGIFactory2(
        0,
        __uuidof<IDXGIFactory7>(),
        (void**)spDxgiFactory.GetAddressOf());    

    D3D_FEATURE_LEVEL featureLevel = D3D_FEATURE_LEVEL.D3D_FEATURE_LEVEL_11_0;
    using ComPtr<ID3D11Device> spWarpDevice = default;
    hr = DirectX.D3D11CreateDevice(
        null,
        D3D_DRIVER_TYPE.D3D_DRIVER_TYPE_WARP,
        HMODULE.NULL,
        (uint)D3D11_CREATE_DEVICE_FLAG.D3D11_CREATE_DEVICE_BGRA_SUPPORT,
        &featureLevel,
        1,
        D3D11.D3D11_SDK_VERSION,
        spWarpDevice.GetAddressOf(),
        null,
        null);
    
    using ComPtr<IPresentationFactory> spPresentationFactory = default;
    hr = CreatePresentationFactory(
        (IUnknown*)spWarpDevice.Get(),
        __uuidof<IPresentationFactory>(),
        (void**)spPresentationFactory.GetAddressOf());
        
    bool isPresentationSupported = (BOOL)spPresentationFactory.Get()->IsPresentationSupported();
    bool isPresentationSupportedWithIFlip = (BOOL)spPresentationFactory.Get()->IsPresentationSupportedWithIndependentFlip();
    
    (isPresentationSupported, isPresentationSupportedWithIFlip).Dump();
}
```

PR #417, opened 2025-11-29
[IMPL] Support `IDxbcConverter` from `dxilconv.dll`

### LINKED ISSUE #416: `IDxbcConverter` Support
#### Description

https://github.com/microsoft/DirectXShaderCompiler/blob/fa3fe957ea9656df75ce683c47fb2892151e4732/projects/dxilconv/include/DxbcConverter.h

#### Rationale

This API is useful for updating DXBC shaders to DXIL.

#### Proposed API

* `DxcCreateInstance`
* `DxcCreateInstance2`
* `IDxbcConverter`

#### Drawbacks

Obviously, this increases the maintenance cost of the repository.

#### Alternatives

I currently have hand-written bindings in my project.

#### Other thoughts

I am willing to make a pull request.

ds5678:
This implements support for `IDxbcConverter` from `dxilconv.dll` in `C:\Windows\System32` using the header from [DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler).

* I created a new class `DxilConv` so that `DxcCreateInstance` and `DxcCreateInstance2` wouldn't conflict with the ones in `DirectX` for `dxcompiler.dll`. Should I have renamed them to `DxilConvCreateInstance` / `DxilConvCreateInstance2` and put them in the `DirectX` class?
* Generated with ClangSharpPInvokeGenerator 20.1.8.2

Resolves #416

### DIFF
--- a/generation/DirectX/dxilconv/DxbcConverter/DxbcConverter.h
+++ b/generation/DirectX/dxilconv/DxbcConverter/DxbcConverter.h
@@ -1,0 +1,62 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxbcConverter.h                                                           //
+// Copyright (C) Microsoft. All rights reserved.                             //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Provides declarations for the DirectX DXBC to DXIL converter component.   //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef __DXBC_CONVERTER__H__
+#define __DXBC_CONVERTER__H__
+
+#include "..\..\..\TerraFX.h"
+#include <dxcapi.h>
+
+#ifndef _MSC_VER
+extern "C"
+#endif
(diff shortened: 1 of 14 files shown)


generation/DirectX/dxilconv/DxbcConverter/DxbcConverter.h
```diff
@@ -0,0 +1,62 @@
+///////////////////////////////////////////////////////////////////////////////
```
tannergooding: This file is notably in the "wrong" location, it should be somewhere under `generation/include`. It also, at minimum, requires updating the THIRD-PARTY-NOTICES.txt in the solution root and would be beneficial to know where it comes from.

That being said... typically the bindings in TerraFX.Interop.Windows are exclusive to what ships with Windows, the Windows SDK, or the DirectX Agility SDK. This header looks to be part of none of those and so is likely a "pit of failure" for users as it won't be usable "out of the box". The fact that it also creates a separate class to avoid conflicting with in-box APIs is problematic as well, because it creates a deviation from what users expect and where they're looking for APIs to be exposed.

This seems like something that should probably be its own repo/project, because its explicitly disconnected from what the DirectX SDK ships

